Issue #24, opened 2026-01-13
Schema for Newton Collision Library Hydroelastic and SDF Contacts


preist-nvidia:
### Is this a new feature, an improvement, or a change to existing functionality?

New Feature

### Feature Description

We need USD schema support for the new contact models such that they can be parsed in Newton, IsaacLab, etc.

### Code of Conduct

- [x] I agree to follow newton-usd-schemas Code of Conduct
- [x] I have searched the [open issues](https://github.com/newton-physics/newton-usd-schemas/issues) and have found no duplicates

PR #46, opened 2026-04-22
Add NewtonSDFCollisionAPI for SDF and hydroelastic collision configuration

mzamoramora-nvidia:
## Description

Adds `NewtonSDFCollisionAPI`, a codeless API schema extending `NewtonCollisionAPI`. Applying it to a `Gprim` configures SDF-based collision generation for the shape; hydroelastic contact is opt-in on top via `newton:hydroelasticEnabled`.

## Attributes

| Attribute | Type | Default | Effect |
|---|---|---|---|
| `newton:sdfMaxResolution` | int | `64` | Max grid dimension; must be a multiple of 8. |
| `newton:sdfTargetVoxelSize` | float | `-inf` (sentinel) | When set (`> 0`), overrides `sdfMaxResolution`. |
| `newton:sdfNarrowBandInner` | float | `-0.1` m | Inner narrow-band depth. |
| `newton:sdfNarrowBandOuter` | float | `0.1` m | Outer narrow-band depth. |
| `newton:sdfPadding` | float | `-inf` (sentinel) | SDF AABB padding [m]. When unset, consumer's default is used (Newton falls back to `newton:contactGap`). |
| `newton:sdfTextureFormat` | token | `"uint16"` | Subgrid storage (`uint8`/`uint16`/`float32`). |
| `newton:hydroelasticEnabled` | bool | `false` | Hydroelastic opt-in. Both shapes in a contact pair must set true. |
| `newton:hydroelasticStiffness` | float | `1e10` N/m³ | Contact stiffness coefficient (force = k_eff · area · depth). |

Inherits from `NewtonCollisionAPI`: `newton:contactMargin`, `newton:contactGap` (which itself implicitly applies `PhysicsCollisionAPI`).

## Design notes

- **Applicability**: restricted to `Gprim` via `apiSchemaCanOnlyApplyTo`, matching `NewtonCollisionAPI` and `NewtonMeshCollisionAPI`.
- **Opt-in via schema application**: applying `NewtonSDFCollisionAPI` is the canonical signal that SDF generation is configured for the shape — no separate enable bool, matching the pattern of the other applied collision APIs.
- **Hydroelastic folded into SDF**: hydroelastic contact uses the same SDF representation, so its configuration lives on the same applied API. Activation is gated by `newton:hydroelasticEnabled` (default `false`); authoring `newton:hydroelasticStiffness` alone is a material parameter that does not flip hydroelastic on.
- **`newton:sdfPadding` vs `newton:contactMargin`**: `sdfPadding` extends the SDF's AABB so distance queries remain valid past the surface; `contactMargin` inflates the contact surface itself for collision detection. Different concepts — hence the deliberately distinct names ("padding" vs "margin") to avoid the CSS-style confusion.
- **`-inf` sentinel convention**: matches `newton:contactGap`. Hard limits describe the user-meaningful range; the docstring documents the sentinel and notes that the consumer (e.g. Newton) chooses the fallback.
- **`hydroelasticStiffness` vs Drake's hydroelastic modulus**: Newton's parameter has units N/m³ (force = k_eff · area · depth). Drake's hydroelastic modulus is in Pa (contact force is a surface integral of a pressure field). The two are not directly mappable — if a Drake-style integration ever lands in Newton, a separate attribute can be added with documented precedence.

Paired with Newton parser PR https://github.com/newton-physics/newton/pull/2533 — both must land together.

## Open questions

- `metersPerUnit` contract — should the schema require meters explicitly, or let importers scale?

## Test plan

```bash
uv run python -m unittest discover -s tests -v
```


newton_usd_schemas/generatedSchema.usda
```diff
@@ -397,6 +397,186 @@ class NewtonCollisionAPI "NewtonCollisionAPI" (
     )
 }
 
+class NewtonSDFCollisionAPI "NewtonSDFCollisionAPI" (
```
andrewkaufman, line 424: Can you move this after `NewtonMeshCollisionAPI` in the file?

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# Unreleased
+
+## Features
+
+- Added `NewtonSDFCollisionAPI`, which extends `NewtonCollisionAPI` with attributes that configure SDF-based collision and optional hydroelastic contact (`newton:sdfEnabled`, `newton:sdfMaxResolution`, `newton:sdfTargetVoxelSize`, `newton:sdfNarrowBand*`, `newton:sdfMargin*`, `newton:sdfTextureFormat`, `newton:hydroelasticEnabled`, `newton:kh`).
+  - Hydroelastic contact is opt-in via `newton:hydroelasticEnabled` (default `false`) on the same API; no separate hydroelastic API is required.
+  - Applying `NewtonSDFCollisionAPI` implicitly applies `NewtonCollisionAPI` and `PhysicsCollisionAPI` as well.
```
andrewkaufman: We don't typically add changelog notes per PR, but since you have it here already 

```suggestion
- Added `NewtonSDFCollisionAPI`, which extends `NewtonCollisionAPI` with attributes that configure SDF-based collision and optional hydroelastic contact.
  - Hydroelastic contact is opt-in via `newton:hydroelasticEnabled` (default `false`) on the same API; no separate hydroelastic API is required.
  - Applying `NewtonSDFCollisionAPI` implicitly applies `NewtonCollisionAPI` and `PhysicsCollisionAPI` as well.
```

newton_usd_schemas/generatedSchema.usda
```diff
@@ -397,6 +397,186 @@ class NewtonCollisionAPI "NewtonCollisionAPI" (
...
+    apiSchemas = ["NewtonCollisionAPI"]
+    doc = """Configures SDF-based collision (and optional hydroelastic contact) for a Newton shape.
```
andrewkaufman:
```suggestion
    doc = """Configures SDF-based collision (and optional hydroelastic contact).
```

newton_usd_schemas/generatedSchema.usda
```diff
@@ -397,6 +397,186 @@ class NewtonCollisionAPI "NewtonCollisionAPI" (
...
+
+    Applying this schema enables signed distance field (SDF) generation for
+    the shape. The SDF is built during model finalization using the configured
+    resolution and narrow band parameters.
```
andrewkaufman:
```suggestion
    the collision shape. The SDF should be built from the base Gprim surface, using the configured resolution and narrow band parameters.
```

newton_usd_schemas/generatedSchema.usda
```diff
@@ -397,6 +397,186 @@ class NewtonCollisionAPI "NewtonCollisionAPI" (
...
+
+    Hydroelastic contact is opt-in via newton:hydroelasticEnabled, and uses
+    the same SDF representation. Not supported for planes and heightfields.
```
andrewkaufman: heightfields aren't a thing in USD (yet) so no need to mention that here. Even for planes, is that generally impossible or is it just that newton runtime does not support it yet? If its the later I think we could skip that from the doc string. If its the former maybe we want to curate the applysTo listing a bit more specifically?

```suggestion
    the same SDF representation. Not supported for planes.
```

newton_usd_schemas/generatedSchema.usda
```diff
@@ -397,6 +397,186 @@ class NewtonCollisionAPI "NewtonCollisionAPI" (
...
+    By applying this schema, the prim also inherits the `NewtonCollisionAPI`
+    and `PhysicsCollisionAPI` schemas."""
+)
+{
+    bool newton:sdfEnabled = true (
+        doc = """Toggles SDF collision for this shape.
+
+        When false, the SDF is not built during finalization even though the
+        API remains applied and all parameters are preserved. This follows
+        the same pattern as physics:rigidBodyEnabled."""
```
andrewkaufman: I still find this a bit odd, to have the extra enabled bool flag... One of our arguments for this separate API was to not need extra toggles for each feature of colliders. But now we keep the toggle anyways. Note MeshCollisionAPI does not have this toggle.

I'm not against it enitrely, just want us to give it a second consideration.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -397,6 +397,186 @@ class NewtonCollisionAPI "NewtonCollisionAPI" (
...
+
+        Range: [8, inf)
+        Units: dimensionless"""
+        limits = {
+            dictionary hard = {
+                int minimum = 8
+            }
+        }
+    )
+    uniform float newton:sdfTargetVoxelSize = 0 (
```
andrewkaufman: How do we feel about 0 rather than `-inf` as the sentinal?

newton_usd_schemas/generatedSchema.usda
```diff
@@ -397,6 +397,186 @@ class NewtonCollisionAPI "NewtonCollisionAPI" (
...
+        doc = """Target voxel size for the sparse SDF grid [m].
```
andrewkaufman: What does `[m]` mean?

newton_usd_schemas/generatedSchema.usda
```diff
@@ -397,6 +397,186 @@ class NewtonCollisionAPI "NewtonCollisionAPI" (
...
+    )
+    uniform token newton:sdfTextureFormat = "uint16" (
+        allowedTokens = ["uint8", "uint16", "float32"]
+        doc = """Subgrid texture storage format for the SDF.
+
+        "uint16" (default) stores subgrid voxels as 16-bit normalized textures
+        (half the memory of "float32"). "float32" stores full-precision values.
+        "uint8" uses 8-bit textures for minimum memory."""
+    )
+    uniform float newton:sdfMargin = 0.05 (
```
andrewkaufman: I think PR #2533 sets `sdf_margin: float | None = None` with a fallback to `shape_gap` when none. If we want to match behaviour, this should be `-inf` (the USD sentinal for none) & we should document that `newton:contactGap` is the fallback.

Though I'm not clear why an unspecified margin falls back to a gap... does it make sense to you?

newton_usd_schemas/generatedSchema.usda
```diff
@@ -397,6 +397,186 @@ class NewtonCollisionAPI "NewtonCollisionAPI" (
...
+        doc = """Hydroelastic contact stiffness coefficient [N/m^3].
+
+        Volumetric stiffness for the hydroelastic contact model. Contact force
+        is computed as: force = kh_eff * area * depth, where kh_eff is the
+        effective stiffness of the two shapes in series (harmonic mean).
+        Higher values produce stiffer contacts. Only consulted when
+        newton:hydroelasticEnabled is true.
+
+        Range: (0, inf)
+        Units: pascal"""
```
andrewkaufman: The docstring says `[N/m^3]` but the `Units:` line says `pascal` (which is N/m² I'm told by my agent). Can you fix whichever is wrong?

newton_usd_schemas/generatedSchema.usda
```diff
@@ -397,6 +397,186 @@ class NewtonCollisionAPI "NewtonCollisionAPI" (
...
+
+        Range: (-inf, 0]
+        Units: distance"""
+        limits = {
+            dictionary soft = {
+                float maximum = 0
+            }
+        }
+    )
+    uniform float newton:sdfNarrowBandInnerFraction = 0 (
```
andrewkaufman: `sdfNarrowBandInnerFraction` uses a **soft** max of 0, but `sdfNarrowBandOuterFraction` uses a **hard** min of 0. Is that intentional?

newton_usd_schemas/generatedSchema.usda
```diff
@@ -397,6 +397,186 @@ class NewtonCollisionAPI "NewtonCollisionAPI" (
...
+    )
+    bool newton:hydroelasticEnabled = false (
+        doc = """Toggles hydroelastic contacts for this shape.
+
+        When true, the shape participates in hydroelastic contact generation;
+        both shapes in a colliding pair must have this attribute true for
+        hydroelastic contacts to be generated. Requires SDF to be configured.
+        Not supported for planes and heightfields."""
+    )
+    float newton:kh = 1e10 (
```
andrewkaufman: `kh` should be expanded to something like `hydroelasticStiffness` or `hydroelasticModulus`. We haven't typically lofted the `k*` names into the schemas, but instead preferred human readable terms.

newton_usd_schemas/plugInfo.json
```diff
@@ -65,10 +65,18 @@
               "bases": [
                 "UsdAPISchemaBase"
               ],
-              "schemaKind": "singleApplyAPI",
```
andrewkaufman, line 68: This PR removes the `["Gprim"]` restriction from `NewtonCollisionAPI`, allowing it to apply to any prim type. Please revert that.

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# Unreleased
+
+## Features
+
+- Added experimental `NewtonSDFCollisionAPI`, which extends `NewtonCollisionAPI` with attributes that configure SDF-based collision and optional hydroelastic contact (`newton:sdfEnabled`, `newton:sdfMaxResolution`, `newton:sdfTargetVoxelSize`, `newton:sdfNarrowBand*`, `newton:sdfMargin*`, `newton:sdfTextureFormat`, `newton:hydroelasticEnabled`, `newton:kh`).
```
andrewkaufman: Why is it experimental?

newton_usd_schemas/generatedSchema.usda
```diff
@@ -397,6 +397,191 @@ class NewtonCollisionAPI "NewtonCollisionAPI" (
     )
 }
 
+class NewtonSDFCollisionAPI "NewtonSDFCollisionAPI" (
+    apiSchemas = ["NewtonCollisionAPI"]
+    doc = """Configures SDF-based collision (and optional hydroelastic contact) for a Newton shape.
+
+    EXPERIMENTAL: NewtonSDFCollisionAPI and its hydroelastic attributes are
```
andrewkaufman: Why did you decide this was needed? Are you expecting the sdf and/or hydroelastic configuration in newton to change?

newton_usd_schemas/generatedSchema.usda
```diff
@@ -397,6 +397,191 @@ class NewtonCollisionAPI "NewtonCollisionAPI" (
...
+            dictionary hard = {
+                float minimum = 0
+            }
+        }
+    )
+    uniform float newton:sdfMarginFraction = 0 (
+        doc = """SDF boundary margin as a fraction of the local-space bounding
+        box diagonal.
+
+        When authored with a non-zero value, overrides newton:sdfMargin at
```
andrewkaufman: It was probably in the design document already, but remind me, what is the reason to have all these `Fraction` alternate forms of the attribute?

newton_usd_schemas/plugInfo.json
```diff
@@ -65,10 +65,18 @@
...
-              "apiSchemaCanOnlyApplyTo": [
-                "Gprim"
-              ]
+              "schemaKind": "singleApplyAPI"
+            },
+            "NewtonPhysicsSDFCollisionAPI": {
```
andrewkaufman, line 87: You should add a `apiSchemaCanOnlyApplyTo` listing for `NewtonPhysicsSDFCollisionAPI` as well. It could be Grpim as above or we could get more restrictive (e.g. you said planes aren't allowed)

tests/test_collision.py
```diff
@@ -32,10 +32,14 @@ def test_api_application(self):
...
         self.assertTrue(self.prim.HasAttribute("newton:contactMargin"))  # from NewtonCollisionAPI
         self.assertTrue(self.prim.HasAttribute("newton:contactGap"))  # from NewtonCollisionAPI
+        # SDF/hydro attrs should NOT be present on bare NewtonCollisionAPI
+        self.assertFalse(self.prim.HasAttribute("newton:sdfMaxResolution"))
+        self.assertFalse(self.prim.HasAttribute("newton:kh"))
 
-    def test_api_limitations(self):
-        xform: Usd.Prim = UsdGeom.Xform.Define(self.stage, "/InvalidType").GetPrim()
-        self.assertFalse(xform.CanApplyAPI("NewtonCollisionAPI"))
+    def test_api_applicability(self):
```
andrewkaufman: What was the reason to change this?

tests/test_collision.py
```diff
@@ -78,6 +82,218 @@ def test_contact_gap(self):
...
+        self.assertTrue(self.prim.HasAttribute("newton:hydroelasticEnabled"))
+        self.assertTrue(self.prim.HasAttribute("newton:kh"))
+        # Inherited from NewtonCollisionAPI
+        self.assertTrue(self.prim.HasAttribute("newton:contactMargin"))
+        self.assertTrue(self.prim.HasAttribute("newton:contactGap"))
+
+    def test_api_applicability(self):
+        # Matches PhysicsCollisionAPI: applicable to any prim type.
+        xform: Usd.Prim = UsdGeom.Xform.Define(self.stage, "/AnyPrim").GetPrim()
+        self.assertTrue(xform.CanApplyAPI("NewtonSDFCollisionAPI"))
```
andrewkaufman: This should assert false

tests/test_collision.py
```diff
@@ -32,10 +32,14 @@ def test_api_application(self):
...
+        self.assertFalse(self.prim.HasAttribute("newton:sdfMaxResolution"))
+        self.assertFalse(self.prim.HasAttribute("newton:kh"))
 
-    def test_api_limitations(self):
-        xform: Usd.Prim = UsdGeom.Xform.Define(self.stage, "/InvalidType").GetPrim()
-        self.assertFalse(xform.CanApplyAPI("NewtonCollisionAPI"))
+    def test_api_applicability(self):
+        # Matches PhysicsCollisionAPI: applicable to any prim type.
+        xform: Usd.Prim = UsdGeom.Xform.Define(self.stage, "/AnyPrim").GetPrim()
+        self.assertTrue(xform.CanApplyAPI("NewtonCollisionAPI"))
```
andrewkaufman: This test is wrong now, it should not apply to an xform

tests/test_collision.py
```diff
@@ -78,6 +82,218 @@ def test_contact_gap(self):
...
+        self.assertEqual(plug_type.typeName, "NewtonPhysicsSDFCollisionAPI")
+        schema_type = Usd.SchemaRegistry().GetSchemaTypeName("NewtonPhysicsSDFCollisionAPI")
+        self.assertEqual(schema_type, "NewtonSDFCollisionAPI")
+
+    def test_api_application(self):
+        self.prim.ApplyAPI("NewtonSDFCollisionAPI")
+        # Should inherit NewtonCollisionAPI and PhysicsCollisionAPI
+        self.assertTrue(self.prim.HasAPI("NewtonSDFCollisionAPI"))
+        self.assertTrue(self.prim.HasAPI("NewtonCollisionAPI"))
+        self.assertTrue(self.prim.HasAPI("PhysicsCollisionAPI"))
```
andrewkaufman, line 100: Lets also assertFalse for MeshCollisionAPI just to show they are not in the same stack

newton_usd_schemas/generatedSchema.usda
mzamoramora-nvidia: It makes it easier to migrate assets from PhysX as they use fractional properties. The fractional properties are scale invariant and that helps to have single value that works well when rescaling assets for domain randomization.

newton_usd_schemas/generatedSchema.usda
mzamoramora-nvidia: `[m]` stands for meters but the existing schemas do not really document units in the docstring, so I removed the mentions of `[m]` in the comments in this commit bb3a025cc7405bec04fb44979dfebd8fff2aeabd

newton_usd_schemas/generatedSchema.usda
mzamoramora-nvidia: Removed the `Not supported` commenst in  20eebba2a21b95338729872eabbc1e779ba4a6a4 Drake does support hydroelastic planes and we might support that in Newton in the future.

tests/test_collision.py
mzamoramora-nvidia: Fixed in 89004556bb69

newton_usd_schemas/plugInfo.json
mzamoramora-nvidia: Reverted in 89004556bb69

tests/test_collision.py
mzamoramora-nvidia: Behavior restored. This was a leftover from earlier experiments on applying the schema to Xforms.

newton_usd_schemas/plugInfo.json
mzamoramora-nvidia: Added `"apiSchemaCanOnlyApplyTo": ["Gprim"]`. Drake supports hydroelastic planes and we could do it in the future as well.

tests/test_collision.py
mzamoramora-nvidia: Rename to `test_api_limitations` and asserting false now.

tests/test_collision.py
mzamoramora-nvidia: Done in 71539c7419

newton_usd_schemas/generatedSchema.usda
mzamoramora-nvidia: Fixed the units in ce941c5f70987e02abbb587260c939f9f460c710

newton_usd_schemas/generatedSchema.usda
mzamoramora-nvidia: Rename to  `hydroelasticStiffness` in   ce941c5f70987e02abbb587260c939f9f460c710

newton_usd_schemas/generatedSchema.usda
mzamoramora-nvidia: `newton:hydroelasticStiffness` is a good example of why it might be good to have an experimental marker on the SDF/hydroelastic surface.

Newton's hydroelastic contact computes contact force as:
```
F = k_eff · area · depth
```
this is why `newton:hydroelasticStiffness` has units of N/m³.

Drake's hydroelastic model computes forces as an integral over the iso-pressure surface:
```
F = ∫ E · e(x) · n(x) dA
```
where `E` is the hydroelastic modulus (Pa), `e(x) ∈ [0, 1]` is a dimensionless penetration extent on the contact surface, and `n(x)` is the surface normal at `x`.

Not sure if there can be an easy mapping from `newton:hydroelasticStiffness` to `E`. So, eventually we might need to add a Drake-style modulus  (`E`) attribute.

CHANGELOG.md
mzamoramora-nvidia: See https://github.com/newton-physics/newton-usd-schemas/pull/46#discussion_r3273807255 . Maybe only some parts of the Schema should be marked as experimental?

newton_usd_schemas/generatedSchema.usda
mzamoramora-nvidia: I changed the sentinels to be `-inf` in 534951ed0887f29af508f128e36b985e80178ea2 for:

- `newton:sdfTargetVoxelSize`, 
- `newton:sdfMargin`, 
- `newton:sdfMarginFraction`,  
- `newton:sdfNarrowBandOuterFraction`
- `newton:sdfNarrowBandInnerFraction`.

newton_usd_schemas/generatedSchema.usda
mzamoramora-nvidia: Sentinel was changed to be `-inf` in https://github.com/newton-physics/newton-usd-schemas/commit/534951ed0887f29af508f128e36b985e80178ea2

About defaulting to gap, Newton main passes `shape_gap` as the SDF generation margin (see [builder.py#L10207](https://github.com/newton-physics/newton/blob/2692f3d4/newton/_src/sim/builder.py#L10207)) , which is documented as ["extra AABB padding"](https://github.com/newton-physics/newton/blob/2692f3d4/newton/_src/geometry/sdf_texture.py#L1396).

PR #2533 just makes the SDF margin overridable and keeps gap as the default so existing setups don't change behavior.

The rationale: broad-phase considers pairs within `gap` of the surface; the SDF only gives accurate distances inside its own margin. Tying `sdf_margin` to `gap` when unset guarantees the SDF covers the broad-phase reach. Maybe @nvtw has some opinions about this.

newton_usd_schemas/generatedSchema.usda
mzamoramora-nvidia: Yes. I will do it after the other conversations are resolved. I'm concerned that moving it now will make the conversations harder to track.

newton_usd_schemas/generatedSchema.usda
mzamoramora-nvidia: `sdfNarrowBandInnerFraction` now has a hard max `0` in 534951e  and the follow up commit cef3f09  added `hard min = -1` to forbid bands wider than the bbox.  But I'm also ok with leaving the range as ` (-inf, 0) `

newton_usd_schemas/generatedSchema.usda
mzamoramora-nvidia: Good point.  I decided to remove the bool in 67fd5e9c1608e0123d39279fadb82f1ff3927d4b for simplicity and to keep consistency with `MeshCollisionAPI`.

mzamoramora-nvidia:
@andrewkaufman thanks for the great feedback. I think this is ready for a second round of reviews.

newton_usd_schemas/generatedSchema.usda
andrewkaufman: We discussed in the working group & don't think the explicit experimental paragraph is needed here. Even if Drake's model requires a new formulation later on, we can address that by adding a new attribute & documenting a precedence ordering (like you did with some others here)

newton_usd_schemas/generatedSchema.usda
andrewkaufman: From the WG meeting, we found it a bit odd that these are needed on the schema but _not on the ShapeConfig_. Does a python user creating a robot entirely in code not also have the same rescaling requirements? I guess they are expected to handle it in their own code?

Basically, we were wondering if it'd be reasonable to leave out the `*Fraction` attributes for now & add them later if user feedback shows the demand? Its not a breaking change to add them later, but it would be a breaking change to remove them later.

@nvtw for thoughts as well 🙏

CHANGELOG.md
andrewkaufman: @mzamoramora-nvidia did you see this one?

tests/test_collision.py
```diff
@@ -78,6 +81,205 @@ def test_contact_gap(self):
...
+        self.assertTrue(self.prim.HasAttribute("newton:hydroelasticEnabled"))
+        self.assertTrue(self.prim.HasAttribute("newton:hydroelasticStiffness"))
+        # Inherited from NewtonCollisionAPI
+        self.assertTrue(self.prim.HasAttribute("newton:contactMargin"))
+        self.assertTrue(self.prim.HasAttribute("newton:contactGap"))
+
+    def test_api_limitations(self):
+        xform: Usd.Prim = UsdGeom.Xform.Define(self.stage, "/InvalidType").GetPrim()
+        self.assertFalse(xform.CanApplyAPI("NewtonSDFCollisionAPI"))
+
```
chschuma-disney, line 120: Missing test for `newton:sdfMargin`.

tests/test_collision.py
```diff
@@ -78,6 +81,205 @@ def test_contact_gap(self):
...
+    def test_api_application(self):
+        self.prim.ApplyAPI("NewtonSDFCollisionAPI")
+        # Should inherit NewtonCollisionAPI and PhysicsCollisionAPI
+        self.assertTrue(self.prim.HasAPI("NewtonSDFCollisionAPI"))
+        self.assertTrue(self.prim.HasAPI("NewtonCollisionAPI"))
+        self.assertTrue(self.prim.HasAPI("PhysicsCollisionAPI"))
+        self.assertFalse(self.prim.HasAPI("NewtonMeshCollisionAPI"))
+
+        # SDF attrs
```
chschuma-disney, line 103: Missing check for `newton:sdfMargin`.

newton_usd_schemas/generatedSchema.usda
mzamoramora-nvidia: Removed experimental tags in 0ee058f5c91b416df5a5c68baca1f5f6dd4a24ac

chschuma-disney (changes requested):
This looks mostly good to me, once Andrew's recent comments are addressed and the missing tests are added. I'm also hesitant about the fractional attributes, but might be good to get more feedback on this from devs.

CHANGELOG.md
mzamoramora-nvidia: Yes. I was hoping to update the changelog once at the end, when all other comments were resolved. 
But in the meantime 29209001feb2df640d30d47d5ce1c1567a5b648b upadtes the changelog notes.

tests/test_collision.py
mzamoramora-nvidia: Done in ec6a5d6ed6f2df7d33b116a0a1bd30f228a9521c

tests/test_collision.py
mzamoramora-nvidia: Done in ec6a5d6ed6f2df7d33b116a0a1bd30f228a9521c

newton_usd_schemas/generatedSchema.usda
mzamoramora-nvidia: Agree that we can always add them in the future if needed. So, 6cc8ee9981d988e85bfd3c9d74e16e7673b3c356 removes fractional attributes.

chschuma-disney (dismissed):
Looks good to me, the only thing missing from my side is [moving the schema within the file](https://github.com/newton-physics/newton-usd-schemas/pull/46#discussion_r3274394127).

newton_usd_schemas/generatedSchema.usda
mzamoramora-nvidia: Done in 90d7584dfbc2adeff9dad21a8651d0168add8267

mzamoramora-nvidia:
@andrewkaufman , @chschuma-disney comments should be resolved now.

chschuma-disney (dismissed):

newton_usd_schemas/generatedSchema.usda
```diff
@@ -421,6 +421,126 @@ class NewtonMeshCollisionAPI "NewtonMeshCollisionAPI" (
...
+        doc = """SDF boundary margin.
+
+        Enlarges the SDF boundary beyond the mesh AABB by this distance,
+        enabling precise distance queries slightly outside the geometry.
+
+        This is separate from newton:contactMargin which controls collision
+        detection inflation.
+
+        If `newton:sdfMargin` is set to `-inf`, newton:contactGap is used
+        instead.
```
andrewkaufman: Trying to see if this helps clarify why gap is used (rationale taken from your reply on https://github.com/newton-physics/newton-usd-schemas/pull/46/files/b592852a445108c31b67e3fbb073d0241635442f#r3274334944)

```suggestion
        If `newton:sdfMargin` is set to `-inf`, `newton:contactGap` is used as the SDF margin fallback.
        Since broad-phase collision deterction considers pairs within `gap` of the surface and an SDF
        only provides accurate distances within its own margin, this fallback ensures the SDF covers
        the broad-phase reach.
```

However, when reading that I wonder if it supposed to be `margin+gap` as the fallback since margin inflates the surface & gap is added on top of it? Is the fallback to just gap for SDFs something that got left behind in the larger margin+gap overhaul?

newton_usd_schemas/generatedSchema.usda
```diff
@@ -421,6 +421,126 @@ class NewtonMeshCollisionAPI "NewtonMeshCollisionAPI" (
...
+                float minimum = 0
+            }
+        }
+    )
+    bool newton:hydroelasticEnabled = false (
+        doc = """Toggles hydroelastic contacts for this shape.
+
+        When true, the shape participates in hydroelastic contact generation;
+        both shapes in a colliding pair must have this attribute true for
+        hydroelastic contacts to be generated. Requires SDF to be configured."""
```
andrewkaufman: This feels redundant since its the only way to get the attribute

```suggestion
        hydroelastic contacts to be generated."""
```

newton_usd_schemas/generatedSchema.usda
```diff
@@ -421,6 +421,126 @@ class NewtonMeshCollisionAPI "NewtonMeshCollisionAPI" (
...
+    doc = """Configures SDF-based collision (and optional hydroelastic contact).
+
+    Applying this schema enables signed distance field (SDF) generation for
+    the collision shape. The SDF should be built from the base Gprim surface, using the configured resolution and narrow band parameters.
+
+    Hydroelastic contact is opt-in via newton:hydroelasticEnabled, and uses
+    the same SDF representation.
+
+    By applying this schema, the prim also inherits the `NewtonCollisionAPI`
+    and `PhysicsCollisionAPI` schemas."""
```
andrewkaufman: There is inconsistency with use if backticks (code blocks) in this PR. I'm sure that's true in the rest of the file too. If you feel like addressing it now, I would prefer to have code blocks around attribute names, schema types, and formula in all the doc strings.

However, feel free to ignore & I'll do a broader pass on the whole file in a followup PR after this merges

newton_usd_schemas/generatedSchema.usda
andrewkaufman: @nvtw for comment as well

newton_usd_schemas/generatedSchema.usda
mzamoramora-nvidia: Done in 5cd81be

newton_usd_schemas/generatedSchema.usda
mzamoramora-nvidia: Yes. This is a very good point. The fallback should be `margin + gap`. I've also realized that we need to improve the behavior of the hydroelastics in newton to consider `margin` and `gap`.  I've created this [issue](https://github.com/newton-physics/newton/issues/2947) to track that. But apart from the work needed on the newton side,  I'm wondering:

- Should we rename this to something like `sdfBoundsPadding` or `sdfAabbPadding` ?
- Is mentioning a fallback for `margin+gap` in the schema leaking implementation details?

andrewkaufman: > Is mentioning a fallback for margin+gap in the schema leaking implementation details?

I little bit perhaps, it it is kind of useful for the end user to know this. I never would have guessed myself.

In similar situations we've so far just said e.g

for gap:
> If `newton:contactGap` is set to `-inf`, Newton's builder default should be used instead.

for solver iterations:
> If set to -1, each solver is free to choose an appropriate default.

Happy to do that here too if you think its best.

Re naming, I'll leave that to you & @nvtw to pick whats best on the ShapeConfig & let that inform the name here. If it does end up with margin  in the name & do like some sentance to explain how/why it differs from the other margin.

andrewkaufman:
Closes #24

newton_usd_schemas/generatedSchema.usda
nvtw: Shall we call it sdfPadding? In html, padding and margin are used. And with a word different from the contact margin and gap, we might avoid confusion?
https://fedmentor.dev/posts/padding-margin/
Please feel free to call this a bad idea in case it is - as non-native English speaker, the choice of words is sometimes a bit random.

mzamoramora-nvidia: `sdfPadding` sound good to me. Renamed to `sdfPadding` in  b3170b4.

chschuma-disney (approved):

andrewkaufman (approved):

Issue #25, opened 2026-01-22
Expose the most important `ShapeConfig` properties via schemas


andrewkaufman:
### Is this a new feature, an improvement, or a change to existing functionality?

Improvement

### Feature Description

We support a subset of `ShapeConfig` already via `NewtonCollisonAPI` and `NewtonMaterialAPI`.

We should expand these (and `NewtonMeshCollisionAPI`) to cover the remainder of [ShapeConfig](https://github.com/newton-physics/newton/blob/main/newton/_src/sim/builder.py#L145)

### Additional context or alternatives

See #24 for the SDF & hydroelastic attributes. This issue is to address the remaining gaps after that one lands.

### Code of Conduct

- [x] I agree to follow newton-usd-schemas Code of Conduct
- [x] I have searched the [open issues](https://github.com/newton-physics/newton-usd-schemas/issues) and have found no duplicates

PR #61, opened 2026-05-22
Add NewtonMassAPI

### LINKED ISSUE #4: Add `NewtonInertiaAPI` with inertia tensor
### Is this a new feature, an improvement, or a change to existing functionality?

New Feature

### Feature Description

`NewtonInertiaAPI` is an extension to `UsdPhysicsMassAPI` to allow specifying the 6 valued inertia tensor rather than diagonal intertia & the principal axes.

The main value adds are:
- Avoids eigenvalue decomposition for source data which already represents the tensor values
- Its 6 values (whereas MassAPI requires 7)
- Will be double precesion to avoid lossy conversion

We will use `-inf` as a sentinal value to indicate "no opinion"

If there is an opinion, `physics:diagonalInertia` and `physics:principalAxes` should be ignored. ie. this attribute overrides the others when both are authored.

We need validators to ensure values are in a sensible range.

### Additional context or alternatives
 
Open Questions:
- What's the name? `newton:localInertia`? `newton:inertia`? `newton:inertiaTensor`?
- Do we use `double[]` or `GfMatrix3d`?
  - There is no `Vec6d` type so we need to pick between these. Both require validation.
  - `double[]` requires more documentation about max length (6 values only) & [ordering convention](https://github.com/newton-physics/urdf-usd-converter/blob/main/docs/concept_mapping.md#element-inertia)
  - `GfMatrix3d` requires validation that its a semmetric upper triangular matrix

### Code of Conduct

- [x] I agree to follow newton-usd-schemas Code of Conduct
- [x] I have searched the [open issues](https://github.com/newton-physics/newton-usd-schemas/issues) and have found no duplicates

andrewkaufman:
## Description

Adds a single-apply API schema with three attributes:
- `newton:inertia` (`double[]`): compact 6-element symmetric inertia tensor with higher precision
- `newton:massModel` (`token`): solid vs shell for implicit mass/inertia computation
- `newton:shellThickness` (`float`): wall thickness for shell mode

Restricted to Xformable prims. Should only be applied alongside `PhysicsCollisionAPI` (shape-level) or `PhysicsRigidBodyAPI` (body-level); validated at runtime, not enforceable at the schema level.

Closes #4 and relates to #25 

## Checklist


havess (approved):

chschuma-disney (approved):
Looks good to me :rocket:

PR #62, opened 2026-05-23
Add contact penalty attributes to NewtonCollisionAPI

### LINKED ISSUE #25: Expose the most important `ShapeConfig` properties via schemas

andrewkaufman:
## Description

Add four per-shape contact model coefficients to `NewtonCollisionAPI`:

| Attribute | ShapeConfig | Type | Default | Units |
|---|---|---|---|---|
| `newton:contactStiffness` | `ke` | float | `-inf` | force / distance |
| `newton:contactDamping` | `kd` | float | `-inf` | force * time / distance |
| `newton:contactFrictionStiffness` | `kf` | float | `-inf` | force * time / distance |
| `newton:contactAdhesion` | `ka` | float | `-inf` | distance |

These are penalty-based contact model parameters that tune how the solver resolves contacts on a per-shape basis. Placement on `NewtonCollisionAPI` (rather than `NewtonMaterialAPI`) follows the physics-vs-numerics principle: these are solver tuning parameters, not physical material properties. Two shapes of the same material may need different stiffness values for numerical stability.

All four attributes use the `-inf` sentinel default (matching the `contactGap` precedent) so each engine can substitute its own builder default at parse time, rather than baking Newton's ShapeConfig values (`2500`, `100`, `1000`, `0`) into the shared schema. All four use soft minimum limits of `0`.

Along with #61 closes #25 

## Checklist

### DIFF
--- a/newton_usd_schemas/generatedSchema.usda
+++ b/newton_usd_schemas/generatedSchema.usda
@@ -474,6 +474,60 @@ class NewtonCollisionAPI "NewtonCollisionAPI" (
             }
         }
     )
+    float newton:contactStiffness = 2500 (
+        doc = """Spring constant of the penalty contact model.
+
+        Higher values produce stiffer contacts at the cost of requiring smaller time steps.
+
+        Range: [0, inf)
+        Units: force/distance"""
+        limits = {
+            dictionary soft = {
+                float minimum = 0
+            }
+        }
+    )
+    float newton:contactDamping = 100 (
+        doc = """Damping coefficient of the penalty contact model.
+
+        Dissipates energy during contact to reduce bouncing.
+
+        Range: [0, inf)
+        Units: force * time / distance"""
+        limits = {
+            dictionary soft = {
+                float minimum = 0
+            }
+        }
+    )
+    float newton:contactFrictionStiffness = 1000 (
+        doc = """Stiffness of the tangential friction response.
+
+        At low sliding speeds, friction force equals `kf * velocity` (viscous regime); at high speeds
+        it is clamped to the Coulomb limit `mu * normal_force`.
+
+        Range: [0, inf)
+        Units: force * time / distance"""
+        limits = {
+            dictionary soft = {
+                float minimum = 0
+            }
+        }
+    )
+    float newton:contactAdhesion = 0 (
+        doc = """Contact adhesion distance.
+
+        When two surfaces are within this distance, an attractive force pulls them together.
+        A value of 0 disables adhesion.
+
+        Range: [0, inf)
+        Units: distance"""
+        limits = {
+            dictionary soft = {
+                float minimum = 0
+            }
+        }
+    )
 }
 
 class NewtonMeshCollisionAPI "NewtonMeshCollisionAPI" (
--- a/newton_usd_schemas/generatedSchema.usda
+++ b/newton_usd_schemas/generatedSchema.usda
@@ -504,13 +504,13 @@ class NewtonCollisionAPI "NewtonCollisionAPI" (
             }
         }
     )
-    float newton:contactFrictionStiffness = -inf (
-        doc = """Stiffness of the tangential friction response.
+    float newton:contactFrictionDamping = -inf (
+        doc = """Damping of the tangential friction response.
 
         At low sliding speeds, friction force equals `kf * velocity` (viscous regime); at high speeds
         it is clamped to the Coulomb limit `mu * normal_force`.
 
-        If `newton:contactFrictionStiffness` is set to `-inf`, the engine's default should be used instead.
+        If `newton:contactFrictionDamping` is set to `-inf`, the engine's default should be used instead.
 
         Range: [0, inf)
         Units: force * time / distance"""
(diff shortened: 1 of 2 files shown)


newton_usd_schemas/generatedSchema.usda
```diff
@@ -395,6 +395,60 @@ class NewtonCollisionAPI "NewtonCollisionAPI" (
             }
         }
     )
+    float newton:contactStiffness = 2500 (
```
andrewkaufman: Open question for all four contact penalty attributes: should the schema defaults match the ShapeConfig defaults (2500, 100, 1000, 0 — as currently implemented) or use `-inf` sentinels like `contactGap` to signal "use the builder default"? The ShapeConfig defaults are convenient for standalone USD authoring, but sentinels make easier for one non-opinion to service many engines (i.e. each engine can choose their own defaults)

newton_usd_schemas/generatedSchema.usda
```diff
@@ -395,6 +395,60 @@ class NewtonCollisionAPI "NewtonCollisionAPI" (
...
+
+        Range: [0, inf)
+        Units: force * time / distance"""
+        limits = {
+            dictionary soft = {
+                float minimum = 0
+            }
+        }
+    )
+    float newton:contactFrictionStiffness = 1000 (
```
andrewkaufman: The name `contactFrictionStiffness` is under debate in the working group — this may be renamed before merge.

newton_usd_schemas/generatedSchema.usda
havess: Generally speaking I think it's wise to allow engines to configure their own defaults. Matching the decision for contactGap makes sense, but I'd like another opinion too.

havess (dismissed):

newton_usd_schemas/generatedSchema.usda
chschuma-disney: Seconding this. I'm thinking if there is a default value specified in the schema, there should be a clear definition of how the feature that the parameter controls behaves (mathematically or conceptually). It might be the case that we could come up with a sufficient definition for the current solvers, but it still allows for more flexibility if we allow future solvers to deviate from it. So my vote goes to using sentinel values.

andrewkaufman: Updated in a0815bb — flipped all four to `-inf` sentinel defaults

andrewkaufman: Updated to `-inf` sentinel defaults (commit 85eeb35 in the rebased history), matching the `contactGap` precedent. Doc strings now say "the engine's default should be used instead" rather than baking in Newton's ShapeConfig values. Agreed this is the right call for cross-engine portability.

Also folded in e13f21d to align the existing `contactGap` doc string to the same "the engine's default" wording for consistency.

Leaving the `contactFrictionStiffness` naming thread separate (still pending WG resolution).

chschuma-disney (dismissed):
Looks good to me :rocket:

newton_usd_schemas/generatedSchema.usda
andrewkaufman: This was change to `contactFrictionDamping`

newton_usd_schemas/generatedSchema.usda
```diff
@@ -474,6 +474,68 @@ class NewtonCollisionAPI "NewtonCollisionAPI" (
             }
         }
     )
+    float newton:contactStiffness = -inf (
```
andrewkaufman, line 477: @eric-heiden re-raised the question whether these are better suited on `NewtonMaterialAPI` vs `NewtonCollisionAPI`

newton_usd_schemas/generatedSchema.usda
```diff
@@ -474,6 +474,68 @@ class NewtonCollisionAPI "NewtonCollisionAPI" (
...
+
+        Range: [0, inf)
+        Units: force * time / distance"""
+        limits = {
+            dictionary soft = {
+                float minimum = 0
+            }
+        }
+    )
+    float newton:contactFrictionDamping = -inf (
```
andrewkaufman: This name is still being debated

andrewkaufman: Back to `contactFrictionStiffness`

newton_usd_schemas/generatedSchema.usda
andrewkaufman: Alternate approach implemented in https://github.com/newton-physics/newton-usd-schemas/pull/65

andrewkaufman:
Closed in favor of #65

PR #65, opened 2026-05-29
Add contact response attributes to NewtonMaterialAPI

andrewkaufman:
## Description

Place the four contact response attributes (`contactStiffness`, `contactDamping`, `contactFrictionStiffness`, `contactAdhesion`) on `NewtonMaterialAPI` for describing how a material behaves on contact.

Material-level placement matches how contact behavior is authored in practice — in the MuJoCo menagerie, 45/50 models that author contact parameters do so via shared defaults inherited by many shapes. The four coefficients form a coherent tuning set alongside existing friction/restitution on `PhysicsMaterialAPI`. Per-shape override via `NewtonCollisionAPI` can be added later as a backward-compatible extension if needed.

This is an alternative to PR #62 (which placed these on `NewtonCollisionAPI`), based on WG discussion.

Uses `contactFrictionStiffness` for `kf` (not `contactFrictionDamping`), per SDD Section 5.4.

## Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/newton_usd_schemas/generatedSchema.usda
+++ b/newton_usd_schemas/generatedSchema.usda
@@ -663,7 +663,7 @@ class NewtonMaterialAPI "NewtonMaterialAPI" (
         Controls how rigid or compliant the material feels on contact. Higher values
         produce stiffer contacts at the cost of requiring smaller time steps.
 
-        If `newton:contactStiffness` is set to `-inf`, Newton's builder default should be used instead.
+        If `newton:contactStiffness` is set to `-inf`, the engine's default should be used instead.
 
         Range: [0, inf)
         Units: force/distance"""
@@ -680,7 +680,7 @@ class NewtonMaterialAPI "NewtonMaterialAPI" (
         Controls how much energy the material absorbs during contact. Higher values
         reduce bouncing.
 
-        If `newton:contactDamping` is set to `-inf`, Newton's builder default should be used instead.
+        If `newton:contactDamping` is set to `-inf`, the engine's default should be used instead.
 
         Range: [0, inf)
         Units: force * time / distance"""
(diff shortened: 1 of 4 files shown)


newton_usd_schemas/generatedSchema.usda
```diff
@@ -656,6 +656,74 @@ class NewtonMaterialAPI "NewtonMaterialAPI" (
...
+        }
+    )
+
+    float newton:contactAdhesion = -inf (
+        doc = """Contact adhesion distance.
+
+        When two surfaces are within this distance, an attractive force pulls them together.
+        A value of 0 or less disables adhesion.
+
+        If `newton:contactAdhesion` is set to `-inf`, Newton's builder default should be used instead.
```
havess: tiny nit: For other solvers/ecosystems that might want to parse these attributes it might be less confusing if we omit "Netwon's builder default" in favor of something like "the parser should use a sensible default".

As it's written I almost feel directed to go dig through the newton builder for the default.

havess (dismissed):

newton_usd_schemas/generatedSchema.usda
andrewkaufman: hmm it was supposed to say "the engine's default" I think my agent has gotten a bit less thorough this week..

andrewkaufman: Good catch — updated all four to say "the engine's default" instead. 47351a8

havess (approved):

chschuma-disney (approved):
Looks good to me :rocket:

Milad-Rakhsha:
FYI, in classical (Cundall & Strack style) DEM methods linear spring-dashpot needs `contactFrictionDamping` as well. Hertz-Mindlin variants also include tangential damping too, so we might as well include this too? 
but I guess if no Newton solver is using tangential damping we can get back to this later.

PR #69, opened 2026-05-29
Try out None sentinals

andrewkaufman:
## Description

Replace sentinel defaults (`-inf`, `-1`) with `None` (no fallback) for schema attributes where "unset" means "solver/engine picks its own default."

### Motivation

UsdPhysics established a pattern of using out-of-range constants as sentinels — e.g. `physics:centerOfMass = (-inf, -inf, -inf)` means "compute from geometry." Newton schemas inherited this pattern:

```usda
float newton:limitStiffness = -inf    # "engine picks its default"
uniform int newton:maxSolverIterations = -1  # "solver picks its default"
```

These sentinels create problems:
- **Awkward ranges:** "Range: [-1, inf)" — is -1 a valid iteration count? No.
- **Consumer burden:** must special-case these values before doing math
- **Not self-documenting:** schema doesn't indicate which values are magic vs real

### Solution

Declare these attributes with `= None` (no fallback value). When no opinion exists, `attr.Get()` returns `None`:

```usda
float newton:limitStiffness = None (
    doc = "If unset, the engine's own default stiffness is used."
)
```

This gives us:
- **Clean ranges:** `[0, inf]` — every expressible value is valid
- **Simple unset check:** `attr.Get() is None`
- **Self-documenting:** `None` clearly means "not authored"

### Where we draw the line

We do NOT replace defaults where `inf` flows directly into math and produces correct behavior:

```usda
float newton:integralMax = inf    # clamp(x, -inf, inf) = unbounded range
float newton:maxEffort = inf      # clamp(effort, -inf, inf) = no effort cap
```

Here `inf` is the real intended value (unbounded clamp range), not a sentinel.

**Decision rule:**
- `inf` meaning "use the builder/solver default" where the importer immediately replaces it → sentinel → `None` (e.g. joint `velocityLimit`: importer converts `inf` → builder default of 1e6)
- `-inf` or `-1` meaning "I haven't chosen, engine/solver pick for me" → sentinel → `None`
- `inf` in `clamp(x, -val, val)` where unbounded IS the physics → real value → keep `inf`

### Why explicit `= None` over omitting the fallback entirely

A schema attribute with no `= <value>` declaration at all also returns `None` from `Get()`:

```usda
float newton:limitStiffness (
    doc = "If unset, the engine's own default stiffness is used."
)
```

This is legal and has the same runtime behavior. We prefer the explicit `= None` because:
- **Communicates intent:** tells schema readers that the unset state is deliberate and meaningful, not an oversight or a forgotten default
- **Distinguishes from "TODO":** an attr with no default could mean "we haven't decided yet" — `= None` says "we decided: unset is the correct default state"
- **Grep-friendly:** you can search the schema for `= None` to find all attrs with this pattern

Both approaches produce identical runtime behavior (`attr.Get() is None` when unauthored). The difference is purely communicative — `= None` makes the design decision visible in the schema source.
### Ergonomic tradeoff

To reset an authored attr back to unset, users call `attr.Block()` or `attr.Clear()` — `attr.Set(None)` is not valid for typed attributes. We think this is acceptable vs sentinel values polluting valid ranges.

### Attrs converted to None (13)

| Attr | Was | API |
|------|-----|-----|
| `newton:maxSolverIterations` | `-1` | NewtonSceneAPI |
| `newton:velocityLimit` | `inf` | NewtonJointAPI |
| `newton:limitStiffness` | `-inf` | NewtonJointAPI |
| `newton:limitDamping` | `-inf` | NewtonJointAPI |
| `newton:shellThickness` | `-inf` | NewtonCollisionAPI |
| `newton:contactGap` | `-inf` | NewtonCollisionAPI |
| `newton:maxHullVertices` | `-1` | NewtonCollisionAPI |
| `newton:sdfTargetVoxelSize` | `-inf` | NewtonCollisionAPI |
| `newton:sdfPadding` | `-inf` | NewtonCollisionAPI |
| `newton:contactStiffness` | `-inf` | NewtonMaterialAPI |
| `newton:contactDamping` | `-inf` | NewtonMaterialAPI |
| `newton:contactFrictionGain` | `-inf` | NewtonMaterialAPI |
| `newton:contactAdhesion` | `-inf` | NewtonMaterialAPI |

### Attrs kept as inf (5 actuator attrs)

| Attr | Reason |
|------|--------|
| `newton:integralMax` | `clamp(integral, -inf, inf)` = unbounded |
| `newton:maxEffort` | `clamp(effort, -inf, inf)` = unbounded |
| `newton:maxMotorEffort` | Caps effort-speed envelope |
| `newton:saturationEffort` | Stall torque limit |
| `newton:velocityLimit` (DC motor) | `vel/inf = 0` = no speed dropoff |

### Precedent in upstream USD schemas

Surveyed all Pixar USD domain schemas for sentinel and no-fallback patterns:

**Sentinel defaults (`-inf`, `-1`):**
- **UsdPhysics**: `centerOfMass = (-inf,-inf,-inf)` (compute from geometry), `gravityMagnitude = -inf` (use 9.81), `lowerLimit = -inf` (no limit)
- **UsdLux**: `shadow:distance = -1` (no max distance), `shadow:falloff = -1` (no falloff)

**Real `inf` values (not sentinels):**
- **UsdPhysics**: `breakForce = inf` (unbreakable), `breakTorque = inf`, `drive:maxForce = inf` (unlimited)

**No-fallback pattern (attr declared without any default):**
- **UsdGeom**: `uOrder`, `vOrder` on NurbsPatch — semantics are "must be authored" (no sensible default)
- **UsdVol**: `fieldIndex` — optional, no default

**`= None` pattern:** not used in any upstream Pixar schema. We would be the first.

### Open question for AOUSD

Is there a semantic difference between `= None` and simply omitting the fallback in a schema definition? Both produce identical runtime behavior (`attr.Get() is None`). The existing no-fallback pattern in UsdGeom/UsdVol means "must author" — we want to express something different: "unset is a valid, deliberate state meaning the engine picks its own default." The explicit `= None` communicates this intent at the schema source level.

- Is this distinction recognized by USD tooling, or is it purely a source-level convention?
- Do USD-aware UIs handle None-default attrs gracefully (blank widget vs error state)?
- Would AOUSD WGs consider `= None` as an alternative to sentinel values for future schema attrs?

## Checklist

- [x] Schema changes in `generatedSchema.usda`
- [x] Tests updated (Block/clear tests for all None attrs)
- [x] Linter passing
- [x] AOUSD forum post: https://forum.aousd.org/t/replacing-sentinel-fallback-values-with-none-in-schema-definitions/2973

### DIFF
--- a/newton_usd_schemas/generatedSchema.usda
+++ b/newton_usd_schemas/generatedSchema.usda
@@ -9,7 +9,7 @@ class NewtonSceneAPI "NewtonSceneAPI" (
     uniform int newton:maxSolverIterations = None (
         doc = """Maximum number of iterations of the physics solver.
 
-        If unset, the engine's default should be used instead.
+        If unset, each solver is free to choose an appropriate default.
 
         Range: [1, inf)
         Units: dimensionless"""
--- a/newton_usd_schemas/generatedSchema.usda
+++ b/newton_usd_schemas/generatedSchema.usda
@@ -6,16 +6,16 @@ class NewtonSceneAPI "NewtonSceneAPI" (
     doc = "`NewtonSceneAPI` applies on top of a `PhysicsScene` prim, providing attributes to control a Newton solver."
 )
 {
-    uniform int newton:maxSolverIterations = -1 (
+    uniform int newton:maxSolverIterations = None (
         doc = """Maximum number of iterations of the physics solver.
 
-        If set to -1, each solver is free to choose an appropriate default.
+        If unset, the engine's default should be used instead.
 
-        Range: [-1, inf)
+        Range: [1, inf)
         Units: dimensionless"""
         limits = {
             dictionary hard = {
-                int minimum = -1
+                int minimum = 1
             }
         }
     )
@@ -450,7 +450,7 @@ class NewtonJointAPI "NewtonJointAPI" (
         }
     )
 
-    float newton:limitStiffness = -inf (
+    float newton:limitStiffness = None (
         doc = """Stiffness of the joint limit spring.
 
         When a DOF's position exceeds the range defined by the joint limits,
@@ -462,15 +462,15 @@ class NewtonJointAPI "NewtonJointAPI" (
         constraint (no spring). Note: some solvers may ignore this attribute
         and always enforce hard positional constraints.
 
-        A value of `-inf` means the engine's own default stiffness is used.
+        If unset, the engine's own default stiffness is used.
 
         The value is broadcast to all DOFs of the joint.
 
-        Range: [0, inf] when authored (sentinel `-inf` defers to engine default).
+        Range: [0, inf]
         Units: effort / degrees (angular DOFs) or effort / distance (linear DOFs)."""
     )
 
-    float newton:limitDamping = -inf (
+    float newton:limitDamping = None (
         doc = """Damping of the joint limit spring.
 
         When a DOF's position exceeds the range defined by the joint limits,
@@ -483,11 +483,11 @@ class NewtonJointAPI "NewtonJointAPI" (
         no spring to damp. Note: some solvers may ignore this attribute and
         always enforce hard positional constraints.
 
-        A value of `-inf` means the engine's own default damping is used.
+        If unset, the engine's own default damping is used.
 
         The value is broadcast to all DOFs of the joint.
 
-        Range: [0, inf) when authored (sentinel `-inf` defers to engine default).
+        Range: [0, inf)
         Units: effort * seconds / degrees (angular DOFs) or effort * seconds / distance (linear DOFs)."""
     )
 }
@@ -547,7 +547,7 @@ class NewtonMassAPI "NewtonMassAPI" (
         controlled by `newton:shellThickness`."""
     )
 
-    float newton:shellThickness = -inf (
+    float newton:shellThickness = None (
         doc = """Wall thickness for shell mass and inertia computation,
         measured inward from the outer surface of the shape.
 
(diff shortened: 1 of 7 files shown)


tests/test_mass.py
```diff
@@ -74,13 +73,19 @@ def test_shell_thickness(self):
...
         success = attr.Set(0.002)
         self.assertTrue(success)
         self.assertTrue(attr.HasAuthoredValue())
         self.assertAlmostEqual(attr.Get(), 0.002)
 
+        # to set the default/fallback, we must block instead of Set(None)
+        # which means there will be no authored value
+        attr.Block()
+        self.assertFalse(attr.HasAuthoredValue())
+        self.assertEqual(attr.Get(), None)
```
andrewkaufman, line 87: If we move forward with this, we should add a similar block of test code for all affected attributes.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -456,15 +456,15 @@ class NewtonCollisionAPI "NewtonCollisionAPI" (
             }
         }
     )
-    float newton:contactGap = -inf (
+    float newton:contactGap = None (
```
andrewkaufman, line 598: Should probably double check how this works in various applications... Do USD aware UIs generally handle this well & offer an appropriate widget?

tests/test_mass.py
```diff
@@ -74,13 +73,19 @@ def test_shell_thickness(self):
...
 
         success = attr.Set(0.002)
         self.assertTrue(success)
         self.assertTrue(attr.HasAuthoredValue())
         self.assertAlmostEqual(attr.Get(), 0.002)
 
+        # to set the default/fallback, we must block instead of Set(None)
+        # which means there will be no authored value
+        attr.Block()
```
andrewkaufman, line 85: For me, this is the only downside, the user cannot call `attr.Set(None)` & needs to understand opinion clearing/blocking semantics...

This can be either `attr.Clear()` or `attr.Block()` (the later must be used to prevent opinions from weaker layers affecting the result)

PR #70, opened 2026-06-03
Rod Api and its attachment

lyd405121:
## Make rods parse easier

- Support close chain

https://github.com/user-attachments/assets/622e2798-2a88-47b9-9602-7e38ff73ac64

- Support graph rods

https://github.com/user-attachments/assets/38a9efcf-5a73-4ec2-8c5d-be6a0a47f4a3

- support attachment 

https://github.com/user-attachments/assets/5d001f0d-672a-43e5-86ec-c56f52e56f08

## Things should disscussed

- Use  youngs modulus, possion ratio, armature,which are not included in vbd but useful and more accurate
- Twist param are added for future solvers
- Attachment using direct node index instead of fixed joint which not allowed in newton's add_usd


andrewkaufman:
@lyd405121 you can likely address the linter failure with `uv run poe format` & you can check results locally with `uv run poe lint`

lyd405121:
## Add rod visual api
- Visual api apply to basic curve only
- Need no modifction for visualize
- Only one attribute for simulation and rendering sync

## Static Test
- usdview load and render usda file with current  schema

closed

graph

## IsaacSim
- We can load it with isaacsim too

isaasim

lightwheel-isaaclab

## More work
- Not tested for simulation and rendering fabric array sync.
- After isaaclab3.0 GA release , I will test it
- There are two things should be done
- One is [zero copy](https://github.com/isaac-sim/IsaacSim/discussions/614)
- Another is [rendering issue](https://github.com/isaac-sim/IsaacSim/discussions/630)


## Code and Assets
- I put all the examples in [my branch](https://github.com/lyd405121/newton-usd-schemas/tree/asset)
- It includes simulation and parsing code
- [Another interesting script](https://github.com/lyd405121/newton-usd-schemas/blob/asset/scripts/gen_text_rod.py) is to make any words into rods

newton_usd_schemas/generatedSchema.usda
```diff
@@ -1125,3 +1125,397 @@ class NewtonPositionBasedClampingAPI "NewtonPositionBasedClampingAPI" (
...
+
+        Each node is a vertex in the rod graph. Segment bodies are created between connected
+        node pairs as defined by `newton:edges` (or implicitly for a linear chain)."""
+    )
+
+    float[] newton:radius (
+        doc = """Capsule radii for rod segment bodies, in metres.
+
+        Array length semantics (E = number of edges / segments):
+        - Empty:    the engine uses a default of 0.005 for all E segments.
```
chschuma-disney: Small point, but following the convention of the rest of the schemas, I would refer to the engine's default without specifying the numerical value itself.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -1125,3 +1125,397 @@ class NewtonPositionBasedClampingAPI "NewtonPositionBasedClampingAPI" (
...
+        Each entry references indices into `newton:points`. Each edge creates one capsule body
+        oriented so its local +Z points from node u to node v. The total number of edges E
+        determines the number of rod segments.
+
+        If empty, the engine assumes a linear chain with E = N-1 segments.
+        If non-empty, branching and junction topologies are supported via `add_rod_graph`."""
+    )
+    
+
+    int[] newton:fixedPoints (
```
chschuma-disney: I think it might be a bit confusing to have two different mechanisms for constraining points: this `fixedPoints` attribute and `NewtonRodAttachmentAPI`. I could imagine that both cases could be covered by `NewtonRodAttachmentAPI` if we also specify the type of attachment (to a body, to world, ...), which would also allow us to extend this in to future to other types of attachments.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -1125,3 +1125,397 @@ class NewtonPositionBasedClampingAPI "NewtonPositionBasedClampingAPI" (
...
+            stretchStiffness = E * A / L    [N/m]
+            bendStiffness    = E * I / L    [N·m/rad]
+
+        Explicit per-segment `newton:stretchStiffness` or `newton:bendStiffness` arrays
+        take precedence over the derived values if authored.
+
+        A value of -1 (default sentinel) means the attribute is unset; the engine uses
+        `newton:stretchStiffness` and `newton:bendStiffness` directly.
+
+        Range: (0, inf) when active; -1 as sentinel for "unset"
```
chschuma-disney: Small point, but using `-inf` as the sentinel value would better align with the other schemas.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -1125,3 +1125,397 @@ class NewtonPositionBasedClampingAPI "NewtonPositionBasedClampingAPI" (
...
+            Polar moment:       J = pi * r^4 / 2
+            Segment length:     L
+
+            twistStiffness = G * J / L    [N·m/rad]
+
+        Only meaningful when `newton:youngsModulus` is also set to a positive value.
+
+        A value of -1 (default sentinel) means the attribute is unset.
+
+        Range: (-1, 0.5] when active; -1 as sentinel for "unset"
```
chschuma-disney: I might be mistaken, but I thought that the theoretical bounds for Poisson's ratio were [-1, 0.5], so including -1. Even if it's not useful for this application, having -1 as the "unset" sentinel might be confusing. I'd recommend using `-inf` instead.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -1125,3 +1125,397 @@ class NewtonPositionBasedClampingAPI "NewtonPositionBasedClampingAPI" (
...
+        Reserved for solvers that support torsional constraints (e.g. Cosserat rod models).
+        Solvers without twist support (e.g. VBD) should silently ignore this attribute.
+
+        Array length semantics (E = number of edges / segments):
+        - Empty:    the engine uses a default of 0.0.
+        - Length 1: the single value is broadcast uniformly to all segments.
+        - Length E: each entry is associated with the corresponding segment/edge."""
+    )
+
+    bool newton:closed = false (
```
chschuma-disney: I'm wondering if this is necessary when we have the option to build the same rod by just specifying the edges explicitly.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -1125,3 +1125,397 @@ class NewtonPositionBasedClampingAPI "NewtonPositionBasedClampingAPI" (
...
+        doc = """Whether to connect the last segment back to the first to form a closed loop.
+
+        Only meaningful for linear chain topology (when `newton:edges` is empty).
+        When true, an extra loop-closing joint is added between segment E-1 and segment 0.
+        This joint is intentionally excluded from the articulation tree (articulations must be acyclic).
+
+        Ignored when `newton:edges` is non-empty."""
+    )
+
+    quatf[] newton:quaternions (
```
chschuma-disney: Minor point, but I'd call this `orientations` or some variant of that. `quaternions` just describes the representation, but not the function.

chschuma-disney (commented):
Thanks for putting together this RodAPI proposal! I did a first pass to get the discussion started.

In general, I would favor the schema being a bit more distanced from the actual implementation in Newton, so that it is general enough that it could also represent other types of rods (e.g., ones not represented by connected capsules), and wouldn't need to be updated if the Newton implementation was modified. This mostly concerns the documentation and default values.

Besides that, I think there's a bigger (structural) question that I believe is worth thinking about or discussing. USD already provides some variant of rod geometry through `BasisCurves`, and in an ideal world, these would provide all the geometry that we need, so the `RodAPI` could apply to a curve and just add all physics attributes. In practice, there's obviously the issue with junctions that are not representable by `BasisCurves`. So we're stuck with either defining everything ourselves, or using `BasisCurves` and then adding a mechanism to attach curves to each other to form a junction. If you have any insights into why one is better, it would be great to share them here.

newton_usd_schemas/generatedSchema.usda
lyd405121: Nice! Attachment to the world, means fixed!

newton_usd_schemas/generatedSchema.usda
lyd405121: Make sense

newton_usd_schemas/generatedSchema.usda
lyd405121: ## Reasons
- The reseason is newton's rod api has this attribute, which will make the parser easy to impelment

image

- And deeper, add_rod api most of the time will use wrap_in_articulation=true to avoid orphan joints 
- If not telling it is a closed form, some solver will fail for looped joints articulation
- As I know, only disney's kamino solver support looped joints 

## My thoughts
- Remove the 'closed' attribute, will be more natual to read
- But we should remove the 'closed' parameter for  add_rod api

newton_usd_schemas/generatedSchema.usda
lyd405121: ##  I agree with you. 
- And this is also the parameter name for add_rod api, I just copy it
- I think a better way is using orientations instaed of quaternions, but make notions in the schema doc

newton_usd_schemas/generatedSchema.usda
lyd405121: I will polish this.

lyd405121:
> Thanks for putting together this RodAPI proposal! I did a first pass to get the discussion started.
> 
> In general, I would favor the schema being a bit more distanced from the actual implementation in Newton, so that it is general enough that it could also represent other types of rods (e.g., ones not represented by connected capsules), and wouldn't need to be updated if the Newton implementation was modified. This mostly concerns the documentation and default values.
> 
> Besides that, I think there's a bigger (structural) question that I believe is worth thinking about or discussing. USD already provides some variant of rod geometry through `BasisCurves`, and in an ideal world, these would provide all the geometry that we need, so the `RodAPI` could apply to a curve and just add all physics attributes. In practice, there's obviously the issue with junctions that are not representable by `BasisCurves`. So we're stuck with either defining everything ourselves, or using `BasisCurves` and then adding a mechanism to attach curves to each other to form a junction. If you have any insights into why one is better, it would be great to share them here.

I have thought about this
- The key point is simulation and rendering data should be sperated

## Case
- For performance, we choose to use less points to simulate
- And for visuals, we use more points to make it smooth

https://github.com/user-attachments/assets/576daf4b-552b-4adf-a3f8-aa1dee64cd5e

- If simulation and rendering using some number of points, you can see there is a small gap between segments
- But if we break the rule for equal number of points, we can use a strategy like skinning animation even rigging

newton_usd_schemas/generatedSchema.usda
```diff
@@ -1125,3 +1125,375 @@ class NewtonPositionBasedClampingAPI "NewtonPositionBasedClampingAPI" (
...
+            /visual_branchA (BasisCurves + NewtonRodVisualCurveAPI)  indices=[2,5,6,7]
+            /visual_branchB (BasisCurves + NewtonRodVisualCurveAPI)  indices=[2,8,9]
+
+    Contact material:
+    Rod segment capsules inherit contact properties from the standard USD material binding
+    mechanism. Apply `MaterialBindingAPI` to this prim and bind a `Material` prim carrying
+    `NewtonMaterialAPI` (and optionally `PhysicsMaterialAPI`). The engine reads
+    `newton:contactStiffness`, `newton:contactDamping`, and `newton:contactFrictionGain`
+    from the bound material for all rod-object contacts."""
+)
```
andrewkaufman: I agree with @chschuma-disney, most of the non-physics attributes on `NewtonRodAPI` appear to be re-formulations of `BasisCurve` (e.g. points, radius->widths, orientations->normals, closed->wrap). In general, we aim to re-use existing schema whenever possible, unless there is sufficent justification to deviate.

> The key point is simulation and rendering data should be seperated. If simulation and rendering using some number of points, you can see there is a small gap between segments

This is very common in USD & is still commonly done with BasisCurves. e.g. for hair in film VFX they would use 2 BasisCurve prims, one is a "guide" purpose & has lower resolution (not just less points per curve, but usually also fewer total number of curves). The other BasisCurves is "default" or "render" purpose, is considerably denser, often uses a non-linear basis, and is "skinned" to the simulated curves to inherit & interpolate the motion.

As some additional context, AOUSD has a proposal for adding Deformables to UsdPhysics, and many of the physics attributes on `NewtonRodAPI` are being proposed on `PhysicsCurvesDeformableSimAPI` & `PhysicsCurvesDeformableMaterialAPI`, both of which apply to `BasisCurves`. There is as well a proposed `PhysicsDeformablePoseAPI` to define "geometry embeddings" between graphics prims & deforming physics prims. Given this direction of the specification, I think it makes sense for Newton to follow suite, by using & extending `PhysicsCurvesDeformableSimAPI` and `PhysicsCurvesDeformableMaterialAPI` with a `NewtonCurvesDeformableSimAPI` and/or `NewtonCurvesDeformableMaterialAPI` if we need additional attributes (e.g. the damping attrs or wrapInArticulation).

newton_usd_schemas/generatedSchema.usda
```diff
@@ -1125,3 +1125,375 @@ class NewtonPositionBasedClampingAPI "NewtonPositionBasedClampingAPI" (
...
+        Each entry references indices into `newton:points`. Each edge creates one capsule body
+        oriented so its local +Z points from node u to node v. The total number of edges E
+        determines the number of rod segments.
+
+        Linear rods, closed loops, branching rods, and junction rods are all expressed with
+        explicit edges."""
+    )
+
+
+    float newton:youngsModulus = -inf (
```
andrewkaufman: I don't think newton rods in the runtime expose youngsModulus of poissonRatio currently. How did you decide to add them here? When do you prefer them over authoring the stiffnesses directly?

newton_usd_schemas/generatedSchema.usda
```diff
@@ -1125,3 +1125,375 @@ class NewtonPositionBasedClampingAPI "NewtonPositionBasedClampingAPI" (
...
+        Range: [0, inf)
+        Units: kg·m²"""
+        limits = {
+            dictionary hard = {
+                float minimum = 0
+            }
+        }
+    )
+
+    float[] newton:stretchStiffness (
```
andrewkaufman: In the AOUSD proposal, these stiffnesses are placed on the material prim instead & are inherantly uniform (the single value case). GeomSubsets can then be used to provide different values to different vertices (or segments).

I would like to know how granularly you want to vary stiffness?

If you only need a couple values over the curve, Subsets may be a viable solution. If you need a unique value for each segment, Subsets become a bit burdensome & this `float[]` could act more like a primvar -- an array to override/drive the material value across the surface. That would simplify the documentation (no need to explain the length 1 case, the array is only used when non-uniform values are necessary).

newton_usd_schemas/generatedSchema.usda
```diff
@@ -1125,3 +1125,375 @@ class NewtonPositionBasedClampingAPI "NewtonPositionBasedClampingAPI" (
...
+
+        Added directly to the diagonal of each capsule body's inertia tensor:
+
+            I_effective = I_capsule + armature * Identity
+
+        Improves numerical stability of the articulation solver for thin or low-mass rods
+        where the natural capsule inertia is very small and causes ill-conditioning.
+
+        Range: [0, inf)
+        Units: kg·m²"""
```
andrewkaufman: Please follow the units notation of the other schemas using "distance" and "mass". It is more confusing perhaps, but in keeping with USD style, as the stage metrics might not be in kg & meters.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -1125,3 +1125,375 @@ class NewtonPositionBasedClampingAPI "NewtonPositionBasedClampingAPI" (
...
+        Units: dimensionless"""
+        limits = {
+            dictionary soft = {
+                float minimum = -1
+                float maximum = 0.5
+            }
+        }
+    )
+
+    float newton:armature = 0.0 (
```
andrewkaufman: Does this correspond to armature of the cable joint in newton runtime? Or if not, where does this armature get recorded? When would you decide to set armature rather than mass (on the `PhysicsDeformableBodyAPI` or `masses` on `PhysicsCurvesDeformableSimAPI`)?

newton_usd_schemas/generatedSchema.usda
```diff
@@ -1125,3 +1125,375 @@ class NewtonPositionBasedClampingAPI "NewtonPositionBasedClampingAPI" (
...
+
+        Array length semantics (E = number of edges / segments):
+        - Empty:    the engine uses a default of 1e5 for all E segments.
+        - Length 1: the single value is broadcast uniformly to all E segments.
+        - Length E: each entry corresponds to one segment in edge order.
+
+        Units: N/m"""
+    )
+
+    float[] newton:stretchDamping (
```
andrewkaufman: The AOUSD proposal does not expose damping attributes, so this would make sense to me as a `NewtonCurvesDeformableMaterialAPI` extension to `PhysicsCurvesDeformableMaterialAPI`, possibly with the extra `float[]` here as well, pending outcome of the stiffness subsets discussion

newton_usd_schemas/generatedSchema.usda
```diff
@@ -1125,3 +1125,375 @@ class NewtonPositionBasedClampingAPI "NewtonPositionBasedClampingAPI" (
...
+        Reserved for solvers that support torsional constraints (e.g. Cosserat rod models).
+        Solvers without twist support (e.g. VBD) should silently ignore this attribute.
+
+        Array length semantics (E = number of edges / segments):
+        - Empty:    the engine uses a default of 0.0.
+        - Length 1: the single value is broadcast uniformly to all segments.
+        - Length E: each entry is associated with the corresponding segment/edge."""
+    )
+
+    quatf[] newton:orientations (
```
andrewkaufman: The AOUSD proposal formulates these as `physics:restNormals` on `PhysicsCurvesDeformableSimAPI`

newton_usd_schemas/generatedSchema.usda
```diff
@@ -1125,3 +1125,375 @@ class NewtonPositionBasedClampingAPI "NewtonPositionBasedClampingAPI" (
...
+
+        Length must be 0 or E (number of edges / segments):
+        - Empty:    the engine derives each segment's orientation from the edge direction
+                    vector (newton:points[v] - newton:points[u]), aligning the capsule's
+                    local +Z with that vector.
+        - Length E: each quaternion explicitly sets the rest-pose frame for the corresponding
+                    segment body, overriding the point-cloud-derived orientation."""
+    )
+
+    bool newton:wrapInArticulation = true (
```
andrewkaufman: I'd like to avoid using `wrap` in the name here, just out of worry it conflates with the wrap mode on BasisCurves

newton_usd_schemas/generatedSchema.usda
```diff
@@ -1125,3 +1125,375 @@ class NewtonPositionBasedClampingAPI "NewtonPositionBasedClampingAPI" (
...
+        doc = """Whether to model the rod as an articulation chain.
+
+        When true, the rod segments are connected as a rigid articulation,
```
andrewkaufman: What happens in newton when false?

newton_usd_schemas/generatedSchema.usda
andrewkaufman: Also, do you think this is necessary as a dedicated attribute or could it make sense to apply `NewtonArticulationRootAPI` to this prim instead?

newton_usd_schemas/generatedSchema.usda
```diff
@@ -1125,3 +1125,375 @@ class NewtonPositionBasedClampingAPI "NewtonPositionBasedClampingAPI" (
...
+        enabling efficient constraint solving for stiff rods."""
+    )
+
+    int newton:collisionFilterNeighborDistance = 1 (
```
andrewkaufman: Is this used only in wrapInArticulation mode? If so, & if we decide to go with `NewtonArticulationRootAPI`, then I wonder if this can be re-formulated along with the collision filtering mode on `NewtonArticulationRootAPI`

newton_usd_schemas/generatedSchema.usda
```diff
@@ -1125,3 +1125,375 @@ class NewtonPositionBasedClampingAPI "NewtonPositionBasedClampingAPI" (
...
+        limits = {
+            dictionary hard = {
+                int minimum = 0
+            }
+        }
+    )
+
+}
+
+class NewtonRodAttachmentAPI "NewtonRodAttachmentAPI" (
```
andrewkaufman: I'd like to understand if this can be modelled with `PhysicsAttachment` and `PhysicsElementCollisionFilter` from the AOUSD proposal

newton_usd_schemas/generatedSchema.usda
```diff
@@ -1125,3 +1125,375 @@ class NewtonPositionBasedClampingAPI "NewtonPositionBasedClampingAPI" (
...
+        fighting against penetration forces at the attachment point.
+
+        Set to false only if contact between the rod and the attached body is intentionally
+        needed (e.g. the rod rests on a surface that it is also constrained to)."""
+    )
+}
+
+class NewtonRodVisualCurveAPI "NewtonRodVisualCurveAPI" (
+    doc = """`NewtonRodVisualCurveAPI` marks a `BasisCurves` prim as a visual representation
+    of a Newton rod strand.
```
andrewkaufman: As mentioned on the first comment, the traditional way to mark a graphics prim vs a simulation prim is using `purpose` -- sim prims are "guide" purpose, while graphics prims are any of "default", "render", or "proxy" purpose

newton_usd_schemas/generatedSchema.usda
```diff
@@ -1125,3 +1125,375 @@ class NewtonPositionBasedClampingAPI "NewtonPositionBasedClampingAPI" (
...
+                int[] newton:rodPointIndices = [2, 5, 6, 7]
+            /visual_branchB (BasisCurves + NewtonRodVisualCurveAPI)
+                int[] newton:rodPointIndices = [2, 8, 9]
+
+    For a simple linear chain with N nodes, a single child curve suffices:
+        /visual (BasisCurves + NewtonRodVisualCurveAPI)
+            int[] newton:rodPointIndices = [0, 1, 2, ..., N-1]"""
+)
+{
+    int[] newton:rodPointIndices (
```
andrewkaufman: I'd like to understand if this can be re-formualted using `PhysicsDeformablePoseAPI` from the AOUSD proposal

newton_usd_schemas/generatedSchema.usda
```diff
@@ -1125,3 +1125,375 @@ class NewtonPositionBasedClampingAPI "NewtonPositionBasedClampingAPI" (
...
+    to this Xform for rendering. Each curve represents one connected strand of the rod graph
+    (e.g. trunk, branch A, branch B). The `newton:rodPointIndices` attribute on each curve
+    maps its vertex positions to rod node indices, allowing an engine or tool to synchronize
+    the rendered curve with the simulated rod in real time.
+
+    Example layout for a branching graph:
+        /World/cable (Xform + NewtonRodAPI, 10 nodes)
+            /visual_trunk   (BasisCurves + NewtonRodVisualCurveAPI)  indices=[0,1,2,3,4]
+            /visual_branchA (BasisCurves + NewtonRodVisualCurveAPI)  indices=[2,5,6,7]
+            /visual_branchB (BasisCurves + NewtonRodVisualCurveAPI)  indices=[2,8,9]
```
andrewkaufman: `BasisCurves` have a packed representation, we shouldnt need to split these out into 3 prims just for branching. We can split them out if that's desirable from a content authoring perspective, or if it makes the geomety embeddings via `PhysicsDeformablePoseAPI` easier. Just noting it's not strictly necessary for visualization.

eg I think this is the same rendered result, where each p variable is a Vec3f point position. p2 appears in the points listing 3 times, which is allowed

```
/visual_trunk   (BasisCurves)  curveVertexCounts[5, 4, 3], points=[p0,p1,p2,p3,p4,p2,p5,p6,p7,p2,p8,p9]
```

andrewkaufman (commented):
Thanks [lyd405121](https://github.com/lyd405121)! Sorry it has taken so long to get you a review of this. I was waiting to be able to refer to the upcoming AOUSD proposal for deformable curves, but it has not been officially posted yet. I have instead included some details of that proposal in my comments here & linked you a draft copy on our Newton chat space. @frankchenlw can share it if you're not on the chat.

My main comment is that I think we need to re-align with this AOUSD proposal, as it is likely to get adopted & become part of the UsdPhysics standard. If we find issues that make it difficult to do this for Newton Rods, we should provide that feedback to AOUSD directly (once the proposal is posted) to ensure the standard doesn't contradict our needs.

newton_usd_schemas/generatedSchema.usda
lyd405121: ## Thanks for your time and advice.
- I think I am not doing it the right way 😭
## Question
- I have a question about where can I find the PhysicsCurvesDeformableSimAPI 
- I searched [openusd](https://github.com/aousd/OpenUSD-proposals/tree/main/proposals) and [aousd](https://github.com/aousd/specifications-public/tree/main/core) and google chat, find nothing

newton_usd_schemas/generatedSchema.usda
lyd405121: - Youngs' modulus  and Poisson ratio are more accurate and [easy to get](https://www.matweb.com/)
- And I thought for robotics, measuable is important
- It is ok to remove this, but please  keep the 'twist' parameter, though newton is not using it

newton_usd_schemas/generatedSchema.usda
lyd405121: I will read aousd and update this

newton_usd_schemas/generatedSchema.usda
lyd405121: - This is a legacy value of newton for making the solver stable
- I reiewed newton's api, it may be deprecated in the future
- So I will delete this

newton_usd_schemas/generatedSchema.usda
lyd405121: Got it

newton_usd_schemas/generatedSchema.usda
lyd405121: Good idea, let me spend sometime to figure it out

newton_usd_schemas/generatedSchema.usda
lyd405121: - Emmm， some solver will break down for orphan joints
- Like mujoco and featherstone，most of the time， it has to be true

newton_usd_schemas/generatedSchema.usda
lyd405121: OK，I will look into it

newton_usd_schemas/generatedSchema.usda
andrewkaufman: Yes, sorry it is not yet posted officially (just due to some administration issue), you can find a tmp copy here [TMP AOUSD UsdPhysics Deformables proposal](https://github.com/andrewkaufman/OpenUSD-proposals/blob/50a48457043e46af1bf2f9a685f67aed759150f1/proposals/physics_deformables/wp_deformable_physics_temp_2026_06_12.md)

newton_usd_schemas/generatedSchema.usda
andrewkaufman: Ok thank you, I will provide this feedback on the AOUSD proposal. I believe @jumyungc has a draft PR to support twist in newton as well, and the AOUSD already does have twist

andrewkaufman:
We should also syncronize with https://github.com/newton-physics/newton/pull/3192 & verify that PR is loading rod they way you need it to. Can you please review the rod section of that PR when you have some time @lyd405121 ?

lyd405121:
## Reduced schema
@andrewkaufman @chschuma-disney 

- Now only 6 property remains in NewtonCurvesDeformableMaterialAPI
```json
   newton:youngsModulus 
   newton:poissonRatio
   newton:stretchDamping 
   newton:bendDamping
   newton:twistDamping 
```
## Detail

### How to make a rod graph
- I now represent graph rods with standard `PhysicsAttachment`: each branch is a `BasisCurves` deformable, and junctions between branches are curve-to-curve attachments. The importer scans these attachments before creating rods, groups connected curves, and calls `add_rod_graph` for the whole graph component. Simple unattached curves still use the normal rod path.

### Fixed points
- I use 'PhysicsAttachment' to make segment attach to the world
- The **Kinematic Deformables** and **Mass Distribution** defines the whole curve a static object, not   flexible enough

### Other changes
- Using periodic to represent ‘Closed’ in newton
- Remove armature and wrapInArticulation

## Test
- I regenerate [all the sample files](https://github.com/lyd405121/newton-usd-schemas/tree/asset/asset) and all of them can be parsed , simulated and rendered . The picture and video  I have posted before can be reproduced in with reduced schema

## Question
- I found young's modulus and poisson ratio are already in the aousd standard for 3D deformables, I am curious about why curve as 1D deformable and cloth as 2D, not having it

- I use sub geom to make every segment has different NewtonCurvesDeformableMaterialAPI，get an warning

image

- The simulation seems right

https://github.com/user-attachments/assets/86b26c85-842c-49da-8ee3-5ac48844ebb6

## The [review](https://github.com/newton-physics/newton/pull/3192/)

- Something the loader have not implemented

```json
  PhysicsAttachment for rigidbody or kinematic body
  Curve-to-curve junction graphs
  World / segment attachments
  USD import path for add_rod_graph()
  Per-segment materials via GeomSubset
  Newton damping extensions
  NewtonArticulationRootAPI semantics currently used by our assets
```
- Maybe my [test files](https://github.com/lyd405121/newton-usd-schemas/tree/asset/asset) can help

andrewkaufman:
> I use sub geom to make every segment has different NewtonCurvesDeformableMaterialAPI，get an warning



Can you let me know what version of usdview you are testing in? I just tried in both 24.08 and 25.05 and I did not get those warnings in either.

andrewkaufman:
> Something the loader have not implemented

Can you or @frankchenlw please makes those comments on that PR so the author knows we need those? I already made the PhysicsAttachment comment but not the others.

andrewkaufman:
> I found young's modulus and poisson ratio are already in the aousd standard for 3D deformables, I am curious about why curve as 1D deformable and cloth as 2D, not having it

I've asked for feedback from AOUSD, will let you know when I hear back.

andrewkaufman:
> Fixed points: I use 'PhysicsAttachment' to make segment attach to the world

Do you find this is effective? Or do you think we should consider a per-vertex or per-segment kinematic flag?

andrewkaufman:
> Maybe my [test files](https://github.com/lyd405121/newton-usd-schemas/tree/asset/asset) can help

These are very helpful, I think we should contribute them under https://github.com/newton-physics/newton-assets perhaps. Or maybe they are so simple they can be test cases in newton directly in that PR implementing deformables support.

lyd405121:
> > I use sub geom to make every segment has different NewtonCurvesDeformableMaterialAPI，get an warning
> 
> Can you let me know what version of usdview you are testing in? I just tried in both 24.08 and 25.05 and I did not get those warnings in either.

- Latest version:Version 26.05

lyd405121:
> > Fixed points: I use 'PhysicsAttachment' to make segment attach to the world
> 
> Do you find this is effective? Or do you think we should consider a per-vertex or per-segment kinematic flag?

- It works and only using aousd, other solvers can use it too

lyd405121:
> > Maybe my [test files](https://github.com/lyd405121/newton-usd-schemas/tree/asset/asset) can help
> 
> These are very helpful, I think we should contribute them under https://github.com/newton-physics/newton-assets perhaps. Or maybe they are so simple they can be test cases in newton directly in that PR implementing deformables support.

- Yes, the assets are too simple, and I will leave some comments to the [pr](https://github.com/newton-physics/newton/pull/3192)

lyd405121:
> > Something the loader have not implemented
> 
> Can you or @frankchenlw please makes those comments on that PR so the author knows we need those? I already made the PhysicsAttachment comment but not the others.

done!

andrewkaufman:
I've reproduced your GeomSubset warniing in USD v25.11 as well. It seems related to Hydra 2.0. You can suppress the warning by disabling Hydra 2 (`export USDIMAGINGGL_ENGINE_ENABLE_SCENE_INDEX=0`)

I've logged the issue here:
https://github.com/PixarAnimationStudios/OpenUSD/issues/4121

Note, you probably do want to add the "physics" binding purppose to those assets, so its clear they are not visual materials. If you're using usd-exchange to author assets, we have `usdex.core.bindPhysicsMaterial` to do that. Otherwise, its a trivial 2-liner: https://gitlab-master.nvidia.com/omniverse/usd-exchange/usd-exchange/-/blob/main/source/core/library/PhysicsMaterialAlgo.cpp?ref_type=heads#L155-156

newton_usd_schemas/generatedSchema.usda
```diff
@@ -866,6 +866,92 @@ class NewtonMaterialAPI "NewtonMaterialAPI" (
...
+
+        Range: [-1, 0.5]
+        Units: dimensionless"""
+        limits = {
+            dictionary soft = {
+                float minimum = -1
+                float maximum = 0.5
+            }
+        }
+    )
```
andrewkaufman: My understanding is that `youngsModulus` and `poissonRatio` are going to be added to the AOUSD proposal soon, so we can drop them here for now & just add a note that their inclussion in Newton runtime is a requirement for us.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -866,6 +866,92 @@ class NewtonMaterialAPI "NewtonMaterialAPI" (
...
+
+    float newton:stretchDamping = -inf (
```
andrewkaufman: AOUSD exposes 4 stiffnesses, but we only expose 3 dampings here. Is that intentional?

newton_usd_schemas/generatedSchema.usda
lyd405121: Done

newton_usd_schemas/generatedSchema.usda
lyd405121: - Emmm，I have done this in my repository

image

- It is werid why it not come up in this pr

- Let me check this

newton_usd_schemas/generatedSchema.usda
```diff
@@ -866,6 +866,81 @@ class NewtonMaterialAPI "NewtonMaterialAPI" (
...
+
+    Apply this API to the same `Material` prim that carries
+    `PhysicsCurvesDeformableMaterialAPI`. The standard AOUSD API owns curve thickness
+    and stretch/shear/bend/twist stiffness; this API only stores Newton-specific
+    controls that are not part of the AOUSD curve material schema.
+
+    Newton's curve-deformable runtime requires Young's modulus `E` and Poisson ratio
+    `nu` to be provided by the AOUSD curve material schema once those attributes are
+    standardized;  `E` governs stretch and bend response, while `E` and `nu` determine 
+    the shear modulus `G = E / (2 * (1 + nu))`, which governs shear and twist response.
```
andrewkaufman: The four dampings have no `Units:` line. Every other float attribute in this file carries one, and the direct precedent is `newton:damping` (L408), which uses a dual form because its dimensions differ per DOF type:

```
Units: effort * seconds / degrees (angular DOFs) or effort * seconds / distance (linear DOFs).
```

These four split the same way, so one units string cannot cover all of them.

Author them as the time-derivative siblings of the AOUSD curve stiffnesses, which are stretch/shear = `force` and bend/twist = `force * distance * distance`. Damping is stiffness times time:

```
Units: force * time (stretch, shear) or force * distance * distance * time (bend, twist)
```

Please add that to all four, following this file's convention of "distance"/"force"/"time" rather than named SI units.

On why that is the right target: the AOUSD proposal currently documents all four curve stiffnesses as force/area, which is dimensionally inconsistent for each of them (see PixarAnimationStudios/OpenUSD-proposals#111). AOUSD has indicated the proposal will be updated so that these are true stiffnesses, with units corrected. Newton's importer presently reads them as moduli and multiplies by `A/L`, `A/L`, `I/L`, `J/L` (`newton/_src/utils/import_usd_deformable_cable.py:83-98`); that becomes a plain division by segment rest length `L`, which is a Newton-side change tracked separately. Authoring these dampings in the same convention keeps this schema consistent with the updated proposal and lets the importer apply the identical `L` division to all eight attributes.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -866,6 +866,81 @@ class NewtonMaterialAPI "NewtonMaterialAPI" (
...
+    the shear modulus `G = E / (2 * (1 + nu))`, which governs shear and twist response. 
+    Converting these material moduli into discrete stretch/shear/bend/twist stiffness
+    also depends on cross-sectional area, second and polar moments of area, shear correction, 
+    segment rest length, and the solver's rod discretization."""
+)
+{
+    float newton:stretchDamping = -inf (
+        doc = """Newton stretch damping for curve deformables.
+
+        AOUSD curve materials do not define solver damping. If set to -inf, the importer
```
andrewkaufman: The sentinel wording deviates from this file's standard phrasing. Every other sentinel attribute here names the attribute and refers to the engine, for example L606:

```
If `newton:contactGap` is set to `-inf`, the engine's default should be used instead.
```

Please match that form on all four, naming each attribute and using "engine" rather than "importer":

```
If `newton:stretchDamping` is set to `-inf`, the engine's default should be used instead.
```

"the importer should use its simulator default" also misplaces the responsibility: the sentinel is resolved by whoever builds the model, not specifically by the USD importer.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -866,6 +866,81 @@ class NewtonMaterialAPI "NewtonMaterialAPI" (
...
+    doc = """Newton extension material parameters for AOUSD curve deformables.
+
+    Apply this API to the same `Material` prim that carries
+    `PhysicsCurvesDeformableMaterialAPI`. The standard AOUSD API owns curve thickness
+    and stretch/shear/bend/twist stiffness; this API only stores Newton-specific
+    controls that are not part of the AOUSD curve material schema.
+
+    Newton's curve-deformable runtime requires Young's modulus `E` and Poisson ratio
+    `nu` to be provided by the AOUSD curve material schema once those attributes are
+    standardized;  `E` governs stretch and bend response, while `E` and `nu` determine
```
andrewkaufman: Trailing whitespace on L880, L881 and L883.

newton_usd_schemas/generatedSchema.usda
andrewkaufman: Confirming four is correct, and the author's screenshot matches what the runtime exposes.

`ModelBuilder.add_rod` takes four independent damping arguments (`newton/_src/sim/builder.py:7653/7655/7657/7659`), as do `add_rod_graph` (`:7882-7888`) and `add_joint_cable` (`:5219-5225`). They resolve into four separate `JointDofConfig(target_kd=...)` entries (`:5295-5316`) and reach the VBD solver as four distinct per-DOF coefficients — `JointType.CABLE` is built with `linear_axes=[stretch, shear]` and `angular_axes=[bend, twist]`, and `SolverVBD._init_joint_penalty_k` (`newton/_src/solvers/vbd/solver_vbd.py:1345-1385`) fills four consecutive slots. So four is the right number and none of them collapse.

Two related facts worth recording here:

The defaults are asymmetric. `shear_damping` inherits `stretch_damping` only when **both** `shear_stiffness` and `shear_damping` are unset (`builder.py:5298-5303`); the same condition governs twist inheriting bend (`:5306-5312`). Otherwise it is 0.0. Authoring `newton:stretchDamping` alone therefore changes shear behavior as a side effect, and authoring `newton:shearStiffness` silently switches shear damping from inherited to zero. That is worth a sentence in the class doc.

Nothing consumes these yet. `import_usd_deformable_cable.py:708-721` calls `add_rod` with no damping arguments, so all four are inert until the importer is plumbed. That is newton-physics/newton#3485's scope, not this PR's, but it means these attributes cannot be validated end-to-end yet.

newton_usd_schemas/generatedSchema.usda
andrewkaufman: The note in the class doc satisfies this, and the physics it states is right — `G = E / (2 * (1 + nu))` governing both transverse shear and torsion is the standard Timoshenko rod formulation, and cross-sectional area, second/polar moments, shear correction and segment rest length are the correct list of factors.

Recording the Newton-side gaps it implies, so they don't get lost:

- The cable path does not read `physics:youngsModulus` or `physics:poissonsRatio` at all. `_cable_stiffnesses_from_material` (`newton/_src/utils/import_usd_deformable_cable.py:82-98`) consumes only the four stiffnesses; there are no references to either modulus in that file. Only the volume/tet path reads them (`newton/_src/usd/utils.py:1924-1951`).
- The authoring helper `create_cable_stiffness_from_elastic_moduli` (`newton/_src/utils/cable.py:70-190`) derives `G` but applies it to twist only — its docstring scopes `G` to torsion and `CableStiffness` documents "No `shear` field, by design." Transverse shear currently has no modulus-derived value.
- There is no shear correction coefficient anywhere in the cable path.

So all three items in the doc note are genuinely unimplemented rather than partially present, which is worth knowing when this gets filed against the runtime.

tests/test_rods.py
```diff
@@ -0,0 +1,83 @@
...
+
+        values = {
+            "newton:stretchDamping": 0.01,
+            "newton:shearDamping": 0.02,
+            "newton:bendDamping": 0.03,
+            "newton:twistDamping": 0.04,
+        }
+        for name, value in values.items():
+            attr = self.material.GetAttribute(name)
+            self.assertTrue(attr.Set(value), name)
```
andrewkaufman: Please follow the existing per-attribute test style rather than dict-driven loops.

`tests/test_material.py` is the closest precedent — same sentinel and same soft-limit shape as these four. It gives each attribute its own test (`test_contact_stiffness`, `test_contact_damping`, `test_contact_adhesion`, ...), and each one covers the absence of the attribute before the API is applied, the sentinel fallback, authored-value state, a `Set()` round-trip, and the declared limits:

```python
def test_contact_damping(self):
    self.assertFalse(self.material.HasAttribute("newton:contactDamping"))

    self.material.ApplyAPI("NewtonMaterialAPI")
    attr = self.material.GetAttribute("newton:contactDamping")
    self.assertIsNotNone(attr)
    self.assertFalse(attr.HasAuthoredValue())
    self.assertEqual(attr.Get(), -math.inf)

    success = attr.Set(200.0)
    self.assertTrue(success)
    self.assertTrue(attr.HasAuthoredValue())
    self.assertAlmostEqual(attr.Get(), 200.0)

    if USD_HAS_LIMITS:
        soft = attr.GetSoftLimits()
        self.assertTrue(soft.IsValid())
        self.assertAlmostEqual(soft.GetMinimum(), 0.0)
        self.assertIsNone(soft.GetMaximum())
```

Here the same coverage is spread across four separate tests that each loop over the same four attribute names — `test_api_application`, `test_material_parameter_defaults`, `test_material_parameter_roundtrip` and `test_limits`. Please restructure into `test_stretch_damping`, `test_shear_damping`, `test_bend_damping` and `test_twist_damping`, each self-contained in the form above, keeping `test_api_registered`, `test_api_application` and `test_api_limitations` as they are.

Two gaps that closing this would also fix: nothing currently asserts `HasAuthoredValue()` is false before a value is set (which is what distinguishes a real authored `-inf` from the fallback), and `test_material_parameter_roundtrip` does not assert the attribute was absent prior to `ApplyAPI`.

andrewkaufman (commented):
Thanks @lyd405121, some last comments to address then I think we can get this merged. Once its released I'll update the newton parsing side & ask you to test that PR

newton_usd_schemas/generatedSchema.usda
jumyungc: Just FYI, I’m currently adding per-joint stiffness assignment in Newton. One relevant detail: the discretization length is the joint’s local dual rest length, L_joint = 0.5 * (L_parent + L_child).

newton_usd_schemas/generatedSchema.usda
jumyungc: Just sharing the full context: previously, stretch/shear and bend/twist were handled together and shared the same stiffness. Now that we have four independent controls, users who set only stretch and bend stiffness, as before, get backward-compatible defaults: shear inherits stretch stiffness, and twist inherits bend stiffness. This fallback is Newton-specific and can be reconsidered of course via the deprecation process if independent defaults would be better.

chschuma-disney (commented):
I'm seconding Andrew's comments. Other than that, the PR looks good to me.

newton_usd_schemas/generatedSchema.usda
lyd405121: remove blank space at the end of lines.done

newton_usd_schemas/generatedSchema.usda
lyd405121: - Add unit for damping
- Add a per-joint doc
Done

newton_usd_schemas/generatedSchema.usda
lyd405121: - Use similar description of damping like other property
Done

tests/test_rods.py
lyd405121: - Writing test code for every property instead of in a big loop
- Done

newton_usd_schemas/generatedSchema.usda
```diff
@@ -866,6 +866,102 @@ class NewtonMaterialAPI "NewtonMaterialAPI" (
...
+    the stretch damping for shear. Likewise, if both `physics:twistStiffness` and
+    `newton:twistDamping` are unset, Newton uses the bend damping for twist. If either
+    stiffness is authored without its corresponding damping, that mode instead uses the
+    engine's zero-damping default."""
+)
+{
+    float newton:stretchDamping = -inf (
+        doc = """Newton stretch damping for curve deformables.
+
+        AOUSD curve materials do not define solver damping.
```
chschuma-disney: This comment (and the others for the other attributes) seems a bit superfluous.

chschuma-disney (dismissed):
The schema looks good to me, just had a small comment that doesn't necessarily have to be addressed.

newton_usd_schemas/generatedSchema.usda
lyd405121: Done. Clean up all four description of "AOUSD curve materials do not define solver damping."

chschuma-disney (dismissed):

newton_usd_schemas/generatedSchema.usda
```diff
@@ -866,6 +866,94 @@ class NewtonMaterialAPI "NewtonMaterialAPI" (
...
+        L_joint = 0.5 * (L_parent + L_child)
+
+    where `L_parent` and `L_child` are the adjacent segments' rest lengths. The corresponding
+    per-joint stiffness and damping coefficients are divided by `L_joint`.
+
+    If both `physics:shearStiffness` and `newton:shearDamping` are unset, Newton uses
+    the stretch damping for shear. Likewise, if both `physics:twistStiffness` and
+    `newton:twistDamping` are unset, Newton uses the bend damping for twist. If either
+    stiffness is authored without its corresponding damping, that mode instead uses the
+    engine's zero-damping default."""
```
andrewkaufman: Two things this doc string asserts changed in the [proposal update](https://github.com/PixarAnimationStudios/OpenUSD-proposals/blob/61d83b54b7efbe97ad2f480de885255cd1e593be/proposals/physics_deformables/wp_deformable_physics.md#L167-L180):

1. `physics:shearStiffness` and `physics:twistStiffness` no longer exist — they are `physics:curvesShearStiffness` and `physics:curvesTwistStiffness`, and thickness is `physics:curvesThickness`.
2. `youngsModulus` and `poissonsRatio` are standardized now, so "once those attributes are standardized" is stale. They are a [volumetric fallback](https://github.com/PixarAnimationStudios/OpenUSD-proposals/blob/61d83b54b7efbe97ad2f480de885255cd1e593be/proposals/physics_deformables/wp_deformable_physics.md#volumetric-fallback-for-structural-stiffnesses) with sentinels (10⁶ Pa, 0.3, 10⁻³ m). The proposal also names curve shear as *transverse* shear.

The suggestion also renames the four `newton:*` attributes to match, and drops two paragraphs:

- `L_joint = 0.5 * (L_parent + L_child)` — per-joint conversion is engine behavior, not schema semantics, and nothing here constrains it.
- The cross-mode damping fallback, per your point: with every stiffness resolvable from `E`/`ν`, "stiffness is unset" is no longer an observable state, so shear borrowing stretch's damping described a gap the volumetric fallback closes. Damping has no equivalent chain — the proposal defines no damping modulus — so each attribute stands alone and an unauthored one takes the engine default. That is what replaces it.

The `E`/`ν` requirement is worth keeping, since that is the load-bearing statement that Newton must honor those inputs. Also matched the file's house form for the opening line (`` `NewtonFooAPI` applies on top of a `Bar` `` plus the inherits sentence), as in `NewtonMaterialAPI` and `NewtonMassAPI`.

```suggestion
    doc = """`NewtonCurvesDeformableMaterialAPI` applies on top of a `Material`, providing extra damping attributes for Newton's curve deformables.

    By applying this schema, the prim also inherits the `PhysicsMaterialAPI` schema.

    Apply this schema alongside `PhysicsCurvesDeformableMaterialAPI`, which provides
    `physics:curvesThickness`, per-mode structural stiffnesses
    `physics:curvesStretchStiffness`, `physics:curvesShearStiffness`,
    `physics:curvesBendStiffness` and `physics:curvesTwistStiffness`, and the isotropic
    fallback pair `physics:youngsModulus` and `physics:poissonsRatio`.
    
    This schema extends a `PhysicsCurvesDeformableMaterialAPI` with corresponding damping
    coefficients for each stiffness.

    Newton requires `physics:youngsModulus` and `physics:poissonsRatio` to be honored for
    curve deformables. Where a stiffness is unauthored it is derived from them and
    `physics:curvesThickness`: `physics:youngsModulus` governs the stretch and bend response,
    while both together give the shear modulus
    `G = physics:youngsModulus / (2 * (1 + physics:poissonsRatio))`, which governs the
    transverse shear and twist response.

    The proposal defines no damping modulus, so damping has no equivalent derivation: each
    damping attribute is independent, and an unauthored one falls back to the engine's
    default rather than to another mode's damping."""
```

newton_usd_schemas/generatedSchema.usda
```diff
@@ -866,6 +866,94 @@ class NewtonMaterialAPI "NewtonMaterialAPI" (
...
+        If `newton:twistDamping` is set to `-inf`, the engine's default should be used instead.
+
+        Range: [0, inf)
+        Units: force * distance * distance * time"""
+        limits = {
+            dictionary soft = {
+                float minimum = 0
+            }
+        }
+    )
```
andrewkaufman: These four should carry the `curves` prefix, to align with the [rename of the stiffnesses they pair with](https://github.com/PixarAnimationStudios/OpenUSD-proposals/blob/61d83b54b7efbe97ad2f480de885255cd1e593be/proposals/physics_deformables/wp_deformable_physics.md#L167-L180). The proposal prefixes properties whose meaning depends on the deformable type; damping qualifies as much as stiffness does.

This also affects the `newton:*` names in the class doc string and in `tests/test_rods.py`.

```suggestion
    float newton:curvesStretchDamping = -inf (
        doc = """Newton stretch damping for curve deformables.

        Damps the stretch mode whose stiffness is `physics:curvesStretchStiffness`.

        If `newton:curvesStretchDamping` is set to `-inf`, the engine's default should be used instead.

        Range: [0, inf)
        Units: force * time"""
        limits = {
            dictionary soft = {
                float minimum = 0
            }
        }
    )

    float newton:curvesShearDamping = -inf (
        doc = """Newton transverse shear damping for curve deformables.

        Damps the shear mode whose stiffness is `physics:curvesShearStiffness`.

        If `newton:curvesShearDamping` is set to `-inf`, the engine's default should be used instead.

        Range: [0, inf)
        Units: force * time"""
        limits = {
            dictionary soft = {
                float minimum = 0
            }
        }
    )

    float newton:curvesBendDamping = -inf (
        doc = """Newton bend damping for curve deformables.

        Damps the bend mode whose stiffness is `physics:curvesBendStiffness`.

        If `newton:curvesBendDamping` is set to `-inf`, the engine's default should be used instead.

        Range: [0, inf)
        Units: force * distance * distance * time"""
        limits = {
            dictionary soft = {
                float minimum = 0
            }
        }
    )

    float newton:curvesTwistDamping = -inf (
        doc = """Newton twist damping for curve deformables.

        Damps the twist mode whose stiffness is `physics:curvesTwistStiffness`.

        If `newton:curvesTwistDamping` is set to `-inf`, the engine's default should be used instead.

        Range: [0, inf)
        Units: force * distance * distance * time"""
        limits = {
            dictionary soft = {
                float minimum = 0
            }
        }
    )
```

tests/test_rods.py
```diff
@@ -0,0 +1,126 @@
...
+        self.material.ApplyAPI("NewtonCurvesDeformableMaterialAPI")
+        self.assertTrue(self.material.HasAPI("PhysicsMaterialAPI"))
+        self.assertTrue(self.material.HasAPI("NewtonCurvesDeformableMaterialAPI"))
+
+        for name in (
+            "newton:stretchDamping",
+            "newton:shearDamping",
+            "newton:bendDamping",
+            "newton:twistDamping",
+        ):
```
andrewkaufman, line 36: If the `curves` prefix rename is taken, these names move with it. The rename touches 13 sites in this file: this tuple, plus the `HasAttribute` / `GetAttribute` pair in each of `test_stretch_damping` (L45, L48), `test_shear_damping` (L65, L68), `test_bend_damping` (L85, L88) and `test_twist_damping` (L105, L108).

Renaming the test methods themselves to `test_curves_stretch_damping` and so on would also keep them aligned with the attributes, matching how [`tests/test_material.py`](https://github.com/newton-physics/newton-usd-schemas/blob/2b95036ff65010aeef1a6a09446a5b87ef480542/tests/test_material.py) names each test after the attribute it covers.

```suggestion
        for name in (
            "newton:curvesStretchDamping",
            "newton:curvesShearDamping",
            "newton:curvesBendDamping",
            "newton:curvesTwistDamping",
        ):
```

andrewkaufman (commented):
Reviewed against the AOUSD deformables proposal as updated in https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/111/commits/61d83b54b7efbe97ad2f480de885255cd1e593be, which lands the per-mode structural stiffnesses with corrected units and adds `youngsModulus`/`poissonsRatio` as a volumetric fallback.

Three suggestions, all button-acceptable, all following from that update. The four damping attributes should take the same `curves` prefix AOUSD applied to the stiffnesses they pair with: these material APIs are designed to stack on a single `Material`, so unprefixed names collide with a future surface-damping API in units that differ per deformable type. The class doc string additionally references two attribute names the proposal has since renamed, and describes `youngsModulus`/`poissonsRatio` as not yet standardized when they now are. The third carries the rename into `tests/test_rods.py`.

newton_usd_schemas/generatedSchema.usda
andrewkaufman: @jumyungc please confirm if you agree with the suggested doc change

jumyungc: This looks good to me. I understand that the updated AOUSD proposal and Newton’s direct API use different fallback semantics - for example, when stretch stiffness is authored but shear stiffness is omitted. I agree that this schema should follow the AOUSD semantics; Newton’s direct API can retain its Newton-specific fallback independently.

andrewkaufman:
@lyd405121 thanks for all your patience & revisions. I am going to merge this once CI passes & so can integrate into newton parsing alongside the next deformable parsing update.

Note I just today learned there may be another revision affecting how curvesThickness is defined, but that is just a doc string here & we can adapt if that lands in the AOUSD proposal.

andrewkaufman (dismissed):

lyd405121:
> @lyd405121 thanks for all your patience & revisions. I am going to merge this once CI passes & so can integrate into newton parsing alongside the next deformable parsing update.
> 
> Note I just today learned there may be another revision affecting how curvesThickness is defined, but that is just a doc string here & we can adapt if that lands in the AOUSD proposal.

- I am glad to help :)

chschuma-disney (dismissed):

andrewkaufman:
@lyd405121 sorry I caused a conflict here by merging another PR. Can you rebase on main & resolve the conflict?

lyd405121:
> @lyd405121 sorry I caused a conflict here by merging another PR. Can you rebase on main & resolve the conflict?

- Pick mpm change, merged without conflict now @andrewkaufman

chschuma-disney (approved):

andrewkaufman (approved):

Issue #30, opened 2026-02-03
Add VBD solver schema


andrewkaufman:
### Is this a new feature, an improvement, or a change to existing functionality?

New Feature

### Feature Description

Newton's VBD solver is important for cable & cloth simulation. We should expose relevant solver parameters on a dedicated `NewtonVBDSceneAPI` schema

### Code of Conduct

- [x] I agree to follow newton-usd-schemas Code of Conduct
- [x] I have searched the [open issues](https://github.com/newton-physics/newton-usd-schemas/issues) and have found no duplicates

Milad-Rakhsha-NV:
## Draft Proposal

1. **Solver scene API, consistent with existing solvers.** `NewtonVbdSceneAPI`
   is a single-apply, codeless API that applies only to `PhysicsScene` and
   lists `NewtonSceneAPI` as a built-in API. This matches the existing
   `NewtonXpbdSceneAPI` and `NewtonKaminoSceneAPI`.

2. **`iterations` is not redefined.** The base `NewtonSceneAPI` already provides
   `newton:maxSolverIterations`. VBD's `iterations` maps onto it, the same way
   `NewtonXpbdSceneAPI` relies on the base attribute rather than adding its own.

3. **Namespacing by parameter group.** Attributes use `newton:vbd:` for common
   parameters and `newton:vbd:particle:` / `newton:vbd:rigid:` for the particle
   and rigid (AVBD) groups, mirroring Kamino's `newton:kamino:padmm:` /
   `newton:kamino:constraints:` sub-grouping. (Open question 1 below asks whether
   to keep the sub-namespaces or flatten to `newton:vbd:`.)

4. **Attributes are included based on whether they describe the simulation or
   the run.** An attribute belongs in the schema when it describes the physical
   scene or changes the solver's simulation behavior, so that it travels with the
   asset and is portable across applications. An attribute is left to
   application-level configuration when it controls how the engine executes
   rather than what is simulated — memory preallocation, backend acceleration,
   algorithm plumbing, and debug output. These do not change the simulated result
   in a portable way and are better set by the runtime.

5. **`-inf` sentinel for "use solver fallback."** Constructor arguments that
   default to `None` and fall back to another parameter (the alpha and beta
   overrides) are represented with a `-inf` default, meaning "inherit the solver
   default." The sentinel keeps the attribute optional: when it is left
   unauthored, the solver chooses the effective value.

## Parameters included in the schema

All attributes are `uniform`.

[...]

PR #71, opened 2026-06-24
added vbd schemas

Milad-Rakhsha-NV:
## Description

Adds `NewtonVbdSceneAPI`, a single-apply API schema that extends `NewtonSceneAPI` with configuration for Newton's VBD (Vertex Block Descent) solver, including its AVBD rigid-body path. Applying it implicitly applies `NewtonSceneAPI`, and it can only be applied to a `PhysicsScene`.

Parameters are grouped by prefix:

- `newton:vbd:` — common solver parameters
- `newton:vbd:particle:` — cloth/particle self-contact parameters
- `newton:vbd:rigid:` — rigid-body AVBD penalty/stabilization parameters

Addresses #30.

```mermaid
graph TD
  NewtonSceneAPI["NewtonSceneAPI"]
  VbdSceneAPI["NewtonVbdSceneAPI"]
  VbdSceneAPI -->|"apiSchemas (inherits)"| NewtonSceneAPI
```

## Design

Attributes are included based on whether they describe the simulation or the run. Parameters that control how the engine executes rather than what is simulated — memory preallocation (contact buffer sizes), backend acceleration toggles (tile solve), and solver-wiring flags (external rigid integration) — are left to application-level configuration and excluded.

`iterations` is not duplicated; it maps to the existing `newton:maxSolverIterations` on `NewtonSceneAPI`.

Override attributes that fall back to another parameter use a `-inf` sentinel meaning "inherit the solver default": `avbdJointAlpha`/`avbdContactAlpha` inherit `avbdAlpha`, and `avbdLinearBeta`/`avbdAngularBeta` inherit `avbdBeta`. These use `soft` limits so the sentinel stays valid.

Defaults, ranges, and units were taken from the VBD/AVBD solver source:

- Ranges mirror the constructor validation (`alpha`/`gamma` in `[0, 1]`; betas, stiffness seeds, ceilings, and damping `>= 0`).
- The penalty ramp is `k = min(k_max, k + beta * |C|)`, so `avbdLinearBeta` has units `force / distance²` and `avbdAngularBeta` has units `torque / radian²`.
- Joint damping enters as `K_eff = k * (1 + kd/dt)`, so `jointLinearKd` and `jointAngularKd` are stiffness-proportional Rayleigh coefficients with units of seconds.

## Attributes

| namespace | attributes |
|---|---|
| `newton:vbd:` | `frictionEpsilon` |
| `newton:vbd:particle:` | `selfContactEnabled`, `selfContactRadius`, `selfContactMargin`, `conservativeBoundRelaxation`, `collisionDetectionInterval`, `edgeParallelEpsilon`, `topologicalContactFilterThreshold`, `restShapeContactExclusionRadius` |
| `newton:vbd:rigid:` | `avbdAlpha`, `avbdJointAlpha`, `avbdContactAlpha`, `avbdBeta`, `avbdLinearBeta`, `avbdAngularBeta`, `avbdGamma`, `contactHistory`, `contactStickMotionEps`, `contactStickFreezeTranslationEps`, `contactStickFreezeAngularEps`, `contactKStart`, `jointLinearKe`, `jointAngularKe`, `jointLinearKStart`, `jointAngularKStart`, `jointLinearKd`, `jointAngularKd` |

## Excluded from the schema 

- `integrate_with_external_rigid_solver` — solver wiring, not a scene property
- `particle_enable_tile_solve` — backend acceleration toggle
- `particle_vertex_contact_buffer_size`, `particle_edge_contact_buffer_size`, `rigid_body_contact_buffer_size`, `rigid_body_particle_contact_buffer_size` — memory preallocation
- `particle_external_vertex_contact_filtering_map`, `particle_external_edge_contact_filtering_map` — element-level collision filtering; belongs to a dedicated element collision filter schema
- `rigid_enable_dahl_friction` — deprecated; controlled by model attributes

## Open questions

- `collisionDetectionInterval` is exposed as a tri-state integer; should the "once" / "twice" / "every n" behavior instead split into a token plus interval?

## Checklist

- [ ] I am familiar with the Contributing Guidelines.


newton_usd_schemas/generatedSchema.usda
```diff
@@ -339,6 +339,387 @@ class "NewtonKaminoSceneAPI" (
...
+                float minimum = 0
+            }
+        }
+    )
+
+    uniform float newton:vbd:rigid:contactKStart = 100.0 (
+        doc = """Body-body and body-particle contact penalty seed for AVBD ramping.
+
+        Inert unless the linear beta (`newton:vbd:rigid:avbdLinearBeta`, or the
+        `newton:vbd:rigid:avbdBeta` fallback) is greater than zero; otherwise the penalty stiffness
```
andrewkaufman: "K" is opaque for someone reading the schema without VBD internals. `contactStiffnessStart` would be clearer. Same for `jointLinearKStart`/`jointAngularKStart`. That said, `Ke` is already established in this schema so `KStart` is at least internally consistent. Up to you — just noting the discoverability gap.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -339,6 +339,387 @@ class "NewtonKaminoSceneAPI" (
...
+        Units: dimensionless"""
+        limits = {
+            dictionary soft = {
+                float minimum = 0
+                float maximum = 1
+            }
+        }
+    )
+
+    uniform float newton:vbd:rigid:avbdBeta = 0.0 (
```
andrewkaufman: The ramp formula (`k = min(k_max, k + beta * |C|)`) is only in the PR description, not here. Consider adding it to the docstring so the schema is self-documenting. Also: `Units: see description` may confuse UX parsers that extract the `Units:` line mechanically — `Units: mixed (feeds both linear and angular ramps)` is clearer.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -339,6 +339,387 @@ class "NewtonKaminoSceneAPI" (
...
+                float maximum = 1
+            }
+        }
+    )
+
+    uniform int newton:vbd:particle:collisionDetectionInterval = 0 (
+        doc = """Frequency of particle self-contact detection.
+
+        A value less than 0 runs detection once before initialization. A value of 0 runs it twice:
+        once before and once after initialization. A value n >= 1 runs it before every n iterations.
```
andrewkaufman: `Range: (-inf, inf)` adds no information since the type is `int`. Either drop the Range line (the tri-state semantics dominate) or narrow it to `Range: [-1, inf)` which reflects the documented minimum useful value.

tests/test_scene_solver_vbd.py
```diff
@@ -0,0 +1,122 @@
...
+        self.assertTrue(count.Set(3))
+        self.assertEqual(count.Get(), 3)
+
+    @unittest.skipUnless(USD_HAS_LIMITS, "USD build does not expose attribute limits")
+    def test_hard_limits(self):
+        self.scene.ApplyAPI("NewtonVbdSceneAPI")
+        # (attribute, minimum, maximum)
+        cases = [
+            ("newton:vbd:frictionEpsilon", 0.0, None),
+            ("newton:vbd:particle:selfContactRadius", 0.0, None),
```
andrewkaufman: This spot-checks 6 of ~15 hard-limited attributes. We need to test all of them. Same for `test_soft_limits_on_sentinel_overrides` — there are 4 sentinel attrs but only 2 are tested.

tests/test_scene_solver_vbd.py
```diff
@@ -0,0 +1,122 @@
...
+                self.assertEqual(float(value), default, name)
+            else:
+                self.assertAlmostEqual(float(value), float(default), places=5, msg=name)
+
+    def test_attribute_set(self):
+        self.scene.ApplyAPI("NewtonVbdSceneAPI")
+        attr = self.scene.GetAttribute("newton:vbd:particle:selfContactRadius")
+        self.assertTrue(attr.Set(0.5))
+        self.assertAlmostEqual(attr.Get(), 0.5, places=5)
+
```
andrewkaufman, line 49: There should be `Set` coverage for all attributes

newton_usd_schemas/_version.py
```diff
@@ -1,4 +1,4 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 The Newton Developers
 # SPDX-License-Identifier: Apache-2.0
 
-__version__ = "0.3.1"
+__version__ = "0.4.0"
```
andrewkaufman: don't bump the version in a feature PR, ill do it at release time

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# 0.4.0
```
andrewkaufman: Typically I don't bother adding changelogs to PRs, I do it all at release time. But since you have a nice writeup already, just cahnge this bit to 
```suggestion
# Unreleased
```

tests/test_scene_solver_vbd.py
```diff
@@ -0,0 +1,122 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 The Newton Developers
```
andrewkaufman:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2026 The Newton Developers
```

andrewkaufman: Requires #72 or cherry-pick that tooling fix in the first commit

tests/test_scene_solver_vbd.py
```diff
@@ -0,0 +1,122 @@
...
+    def test_api_application(self):
+        self.scene.ApplyAPI("NewtonVbdSceneAPI")
+        self.assertTrue(self.scene.HasAPI("NewtonSceneAPI"))
+        self.assertTrue(self.scene.HasAPI("NewtonVbdSceneAPI"))
+
+    def test_api_limitations(self):
+        prim: Usd.Prim = self.stage.DefinePrim("/NotScene", "Xform")
+        self.assertFalse(prim.CanApplyAPI("NewtonVbdSceneAPI"))
+
+    def test_attribute_defaults(self):
```
andrewkaufman: I don't mind particularly, but this is style divergence from the other tests files, which handle everything about one attribute at a time. Is there a reason you went this way? I find it easier to catch omissions when the tests are on a per attribute basis.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -339,6 +339,387 @@ class "NewtonKaminoSceneAPI" (
...
+        Units: dimensionless"""
+        limits = {
+            dictionary hard = {
+                float minimum = 0
+                float maximum = 1
+            }
+        }
+    )
+
+    uniform int newton:vbd:particle:collisionDetectionInterval = 0 (
```
andrewkaufman: > collisionDetectionInterval is exposed as a tri-state integer; should the "once" / "twice" / "every n" behavior instead split into a token plus interval?

I do find tri-state a bit awkward, but if that's what the solver is doing & expects, it seems clear enough from the description :shrug:

andrewkaufman (commented):
Thanks @Milad-Rakhsha-NV, I didn't comment on attribute names & values much, I'll leave that to the VDB experts.

mmichelis:
Thanks for the great effort! All relevant parameters are there for the demos we run in IsaacLab (so far). For the rigid contact, there is another potentially useful boolean to expose: rigid_contact_hard. Do you think this one could make sense?

Milad-Rakhsha-NV:
> Thanks for the great effort! All relevant parameters are there for the demos we run in IsaacLab (so far). For the rigid contact, there is another potentially useful boolean to expose: rigid_contact_hard. Do you think this one could make sense?

Hi Mike we can add it but the authors of VBD was in favor of not exposing it since it will be removed in the future,  @jumyungc for visibility

newton_usd_schemas/generatedSchema.usda
Milad-Rakhsha-NV: I agree, contactStiffnessStart, jointLinearStiffnessStart, jointAngularStiffnessStart, etc are clearer. @jumyungc @AnkaChan would you be fine with these names

newton_usd_schemas/generatedSchema.usda
```diff
@@ -339,6 +339,388 @@ class "NewtonKaminoSceneAPI" (
     )
 }
 
+class "NewtonVbdSceneAPI" (
+    apiSchemas = ["NewtonSceneAPI"]
+    doc = "Provides Newton's VBD (Vertex Block Descent) solver configuration."
+)
+{
+    uniform float newton:vbd:frictionEpsilon = 0.01 (
```
mmacklin: @AnkaChan is this global? Should we key it off the `contact_friction_gain` parameter we added to the contact schema?

newton_usd_schemas/generatedSchema.usda
```diff
@@ -339,6 +339,388 @@ class "NewtonKaminoSceneAPI" (
...
+        Range: [0, inf)
+        Units: distance / seconds"""
+        limits = {
+            dictionary hard = {
+                float minimum = 0
+            }
+        }
+    )
+
+    uniform bool newton:vbd:particle:selfContactEnabled = false (
```
mmacklin: I think a big question here is what to do when we move self contact and triangle based contact out of VBD and into the collision pipeline object.

tests/test_scene_solver_vbd.py
andrewkaufman: Should be fixed if you rebase on main

andrewkaufman:
Fixes #30

jumyungc:
FYI, I added a minimal, starting set of rigid/AVBD params to carry over (https://github.com/newton-physics/newton-usd-schemas/issues/30). It'll need to expand later.

newton_usd_schemas/generatedSchema.usda
Milad-Rakhsha-NV: Thanks Miles I think based on JC's input we will remove this for the initial design and will add it later on if needed

Milad-Rakhsha-NV:
> FYI, I added a minimal, starting set of rigid/AVBD params to carry over (#30). It'll need to expand later.

Thanks JC, will update soon.

jumyungc (commented):
The rigid-body side looks good to me overall - thanks!. 
~~I only added a few small comments to tighten up the doc. 
Just in case, please don’t forget to update the changelog.~~

This needs further testing, but after thinking about it more, we may not need joint_Linear/Angular_Ke/Kd. contactHistory may also be unnecessary. To keep the schema conservative and minimal, I'd suggest not shipping any rigid-specific parameters for now.

newton_usd_schemas/generatedSchema.usda
jumyungc: I’ll mark this as resolved since it is no longer part of the new schema.

jumyungc:
Sorry for the delay! I’ve added a minimal set of rigid-body parameters for VBD’s new compliant ALM path in https://github.com/newton-physics/newton-usd-schemas/issues/30

Issue #75, opened 2026-07-21
Add schemas for authored MPM scenes, particles, and materials


maxkra15:
## Motivation

Newton's implicit MPM solver can currently be configured only through Python/`ModelBuilder`. The Isaac Sim integration needs a portable USD representation that opts authored particle data into MPM explicitly, carries solver/material parameters, and round-trips without procedural particle generation.

Impacted integration/company: NVIDIA Isaac Sim and Isaac Lab's Newton MPM workflows.

This issue is intended to align schema names and field semantics with the code owners before an upstream PR, per `CONTRIBUTING.md`.

## Proposed APIs

- `NewtonMPMSceneAPI`, restricted to `PhysicsScene`, and including `NewtonSceneAPI`.
- `NewtonMPMParticleAPI`, restricted to `Points`.
- `NewtonMPMMaterialAPI`, restricted to `Material` and including `PhysicsMaterialAPI`.

The API names above are working names; feedback on naming and namespace conventions is explicitly requested before merge.

## Proposed representation

`UsdGeom.Points` remains the source of particle positions, velocities, widths (diameters), and ids. Applying `NewtonMPMParticleAPI` is the explicit simulation opt-in, so render-only Points are ignored. The particle API is an attribute-free marker:

- Per-point mass is not MPM-specific and is out of scope for this schema. Importers can derive Newton particle mass from standard `physics:density` and particle width; an authored generic Points-mass representation should be designed separately and aligned with the deformables proposal's `physics:masses` work.
- Newton's `ACTIVE` flag is generic particle participation, so imported particles default active.
- Newton's experimental `PROXY` flag is runtime solver-coupling state.
- Neither flag is authored by this schema.

A physics-purpose material binding selects a Material carrying `NewtonMPMMaterialAPI`. Discrete variation within one Points prim is represented with point-element `GeomSubset` prims in the `materialBind` family, using non-overlapping or partition semantics. Unassigned points inherit the parent binding.

The scene API maps portable `SolverImplicitMPM.Config` fields such as rheology solver order, tolerance, warm-start/collider velocity modes, voxel/grid configuration and capacities, transfer/integration schemes, critical fraction, air drag, and basis selection. Schema fallbacks should match Newton's public config defaults.

[...]

Issue #10, opened 2025-11-26
Add automatable schema validation


andrewkaufman:
### Is this a new feature, an improvement, or a change to existing functionality?

New Feature

### Feature Description

Validation use useful generally, but in particular for #9 and #4, there is intricate requirements across attributes & it would be easy to accidentally author invalid joint state. We should provide automated validation for authored content.

### Additional context or alternatives

We can do this using `omni.asset_validator` (via `usd-exchange[test]`) for now, as `UsdValidation` does not currently support Python rules.

It should _not_ be a strict runtime requirement however, it needs to be opt-in.

### Code of Conduct

- [x] I agree to follow newton-usd-schemas Code of Conduct
- [x] I have searched the [open issues](https://github.com/newton-physics/newton-usd-schemas/issues) and have found no duplicates

andrewkaufman:
`UsdValidation` will support python rules in 26.05+, so we can use that once newton is willing to make 26.05 the minspec.

PR #76, opened 2026-07-22
Add authored MPM USD schemas

maxkra15:
## Description

Adds codeless USD schemas for authoring particle sets and Newton implicit MPM configuration.

- Adds `NewtonMPMSceneAPI`, restricted to `PhysicsScene`, with the stable MPM configuration fields selected during review.
- Adds the generic `NewtonParticleAPI`, restricted to `Points`, using standard `UsdGeom.Points` data and the standard `physics:simulationOwner` relationship.
- Adds `NewtonMPMMaterialAPI`, restricted to `Material`, for MPM constitutive properties. It applies `PhysicsMaterialAPI` but intentionally does not apply the surface-contact-focused `NewtonMaterialAPI`; Young's modulus and Poisson's ratio are omitted here and deferred to the proposed `PhysicsVolumeDeformableMaterialAPI`.
- Represents collider and strain bases with separate type, order, and discontinuity attributes suitable for authoring UIs and validation.
- The fixture demonstrates standard physics-purpose material bindings on both `Points` and point-element `GeomSubset` prims.
- Updates `plugInfo.json` with the appropriate apply-to restrictions.
- Adds a normalized `sand.usda` fixture containing authored particles and two discrete materials.
- Tests registration, `CanApplyAPI`, fallbacks, allowed tokens, limits metadata, simulation ownership, material bindings, and fixture loading.

This schema layer deliberately does not add per-particle mass, runtime particle flags, procedural generation, or emitters. Importers may derive particle mass from particle size and `physics:density`.

Newton importer support and Isaac Sim runtime integration remain separate follow-up work.

Addresses #75.

### Local validation

- `uv run --frozen --group dev poe lint`
- `uv run --frozen --group dev poe test` — 182 tests passed
- `uv build --wheel`

## Checklist


newton_usd_schemas/generatedSchema.usda
```diff
@@ -339,6 +339,217 @@ class "NewtonKaminoSceneAPI" (
     )
 }
 
+class "NewtonMPMSceneAPI" (
```
andrewkaufman, line 342: @gdaviet please give a thorough review of the exposed `NewtonMPMSceneAPI`. We only want to expose the minimal set of "stable" MPM solver attributes to the schema for now. Make sure the names, types, defaults, limits, and doc strings make sense to you please.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -339,6 +339,217 @@ class "NewtonKaminoSceneAPI" (
...
+        Units: dimensionless"""
+        limits = {
+            dictionary hard = {
+                float minimum = 0
+            }
+        }
+    )
+
+    uniform token[] newton:mpm:rheologySolvers = ["auto"] (
+        allowedTokens = ["auto", "gs", "gauss-seidel", "gs-soa", "gauss-seidel-soa", "gs-batched", "gauss-seidel-batched", "jacobi", "cg", "cr", "gmres"]
```
chschuma-disney: If I'm seeing this correctly in the code, "gs", "gs-soa", and "gs-batched" are short names for the Gauss-Seidel solvers, right? In that case, I think it might be worth dropping these from the schema to avoid redundancy.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -339,6 +339,217 @@ class "NewtonKaminoSceneAPI" (
...
+        from the previous time-step position."""
+    )
+
+    uniform float newton:mpm:voxelSize = -inf (
+        doc = """MPM background-grid voxel size.
+
+        A value of `-inf` asks the engine to use its own default. Any authored
+        numeric value is interpreted in the stage's distance units.
+
+        Range: {-inf} or (0, inf)
```
chschuma-disney: The convention so far has been that sentinel values don't need to be added to the range.

The same holds for other instances of this.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -339,6 +339,217 @@ class "NewtonKaminoSceneAPI" (
...
+        }
+    )
+
+    uniform bool newton:mpm:colliderNormalFromSdfGradient = false (
+        doc = """Whether collider normals are computed from the signed-distance-field gradient.
+
+        When false, the solver uses the closest-point normal."""
+    )
+
+    uniform token newton:mpm:colliderBasis = "S2" (
```
chschuma-disney: To make this clearer for the schema, could it make sense to split this into a set of tokens ("Q1", "S2", "pic") and an additional attribute that determines the parameter for "pic"?

newton_usd_schemas/generatedSchema.usda
```diff
@@ -339,6 +339,217 @@ class "NewtonKaminoSceneAPI" (
...
+        doc = """Basis function used to sample colliders.
+
+        Common values are "Q1", "S2", "pic", "pic8", and "pic27". Engines
+        may accept any "picN" token where N is a positive integer, so this
+        attribute intentionally has no closed `allowedTokens` list."""
+    )
+
+    uniform token newton:mpm:strainBasis = "P0" (
```
chschuma-disney: The same here with splitting into tokens and an additional attribute.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -866,6 +1120,201 @@ class NewtonMaterialAPI "NewtonMaterialAPI" (
...
+)
+{
+    float newton:mpm:youngsModulus = -inf (
+        doc = """Young's modulus.
+
+        A value of `-inf` asks the engine to use its own default. Any authored
+        numeric value is interpreted in the stage's derived pressure units.
+
+        Range: {-inf} or (0, inf)
+        Units: pressure"""
```
chschuma-disney: Should probably be `force / (distance * distance)` to match other schemas.

The same for other attributes that use pressure.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -866,6 +1120,201 @@ class NewtonMaterialAPI "NewtonMaterialAPI" (
...
+            dictionary soft = {
+                float minimum = 0
+            }
+        }
+    )
+
+    float newton:mpm:poissonsRatio = 0.3 (
+        doc = """Poisson's ratio for the elastic response.
+
+        Range: (-1, 0.5)
```
chschuma-disney: The USD Physics Deformables proposal defines the range as inclusive, so [-1, 0.5]. Though it might be fine to exclude the limits here if these values are not supported by the solver.

chschuma-disney (commented):
Thanks for getting this started! I took a first pass, skipping some comments that I know Andrew will post later based on the Newton USD meeting.

It would also be great to get @gdaviet to take a pass, to make sure we're not adding anything to the schema that might be up for refactoring/revision.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -765,6 +976,49 @@ class NewtonSiteAPI "NewtonSiteAPI" (
...
+    (or a partition); points not assigned to a subset inherit the parent
+    binding.
+
+    The initial v1 contract resolves one active `NewtonMPMSceneAPI`
+    configuration for all marked `Points` prims in an imported model. Explicit
+    per-set scene relationships and an MPM body/object grouping API such as a
+    future `NewtonMPMBodyAPI` are deferred."""
+)
+{
+}
```
andrewkaufman: Since there is no longer anything MPM specific about this API, its really just a flag that _this `UsdGeom.Points`_ participates in particle physics. I'd suggest we generalize the name so other particle solvers (outside newton runtime) can re-use the API, and add the standard `physics:simulationOwner` relationship for folks who want to target an MPM solver specifically can set that rel to the appropraite `PhysicsScene` prim. We also don't need to re-explain material binding semantics so thorughly.

This makes for a relatively straightforward proposal back to AOUSD at some point.


```suggestion
class NewtonParticleAPI "NewtonParticleAPI" (
    doc = """Applies physics attributes to any `UsdGeomPoints` prim and marks that prim to be driven by a simulation.
    
    Particle positions, velocities, widths, and stable identifiers use the standard `UsdGeomPoints` `points`, `velocities`, `widths` (or `primvars:widths`), and `ids` attributes.
    
    Note that as in `UsdGeomPoints`, width are specified as particle diameter, therefor radius should be computed as `width / 2` per point.
    
    Mass can be derived from particle size and a bound material's `physics:density`.
"""
)
{
    rel physics:simulationOwner (
        doc = """Single `PhysicsScene` that will simulate these particles.
        
        By default this is the first `PhysicsScene` found in the stage during a standard traversal.
        """
    )
}
```

andrewkaufman: Note I added support for `primvars:widths` in that doc string... youre parsing PR needs updating to support that as well. Like all primvars, they can be indexed & they might not be Vertex (per point) interpolation, they could be Contant.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -866,6 +1120,201 @@ class NewtonMaterialAPI "NewtonMaterialAPI" (
...
+    useful beyond MPM. The AOUSD deformables proposal uses
+    `physics:youngsModulus` and `physics:poissonsRatio` for volume-deformable
+    materials, but its simulation geometries do not currently include
+    `UsdGeomPoints`. These attributes remain Newton-namespaced because this API
+    defines Newton's exact implicit-MPM constitutive semantics. When the
+    canonical standard attributes are authored under a recognized standard
+    deformable-material API, Newton importers give them precedence over the
+    corresponding Newton attributes and warn when both authored values
+    conflict. The Newton attributes remain the compatibility path until the
+    standard representation applies directly to Points."""
```
andrewkaufman: Most of this reads as LLM talking points, not a schema documentation. Please reduce it to the minimum requirements for the schema to be self documenting. Products like Kaolin should not be mentioned here. In general our schema docs aim to be newton agnostic when they can as well, and just refer to "the engine", since they are used in newton, mujoco (standalone), and physx. Though I realize that's not as possible for MPM specific things.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -866,6 +1120,201 @@ class NewtonMaterialAPI "NewtonMaterialAPI" (
...
+)
+{
+    float newton:mpm:youngsModulus = -inf (
```
andrewkaufman: I believe this is being added to a base PhysicsMaterialAPI in the deformables proposal. Lets drop it here and use that instead.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -866,6 +1120,201 @@ class NewtonMaterialAPI "NewtonMaterialAPI" (
...
+        Range: {-inf} or (0, inf)
+        Units: pressure"""
+        limits = {
+            dictionary soft = {
+                float minimum = 0
+            }
+        }
+    )
+
+    float newton:mpm:poissonsRatio = 0.3 (
```
andrewkaufman: I believe this is being added to a base PhysicsMaterialAPI in the deformables proposal. Lets drop it here and use that instead.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -866,6 +1120,201 @@ class NewtonMaterialAPI "NewtonMaterialAPI" (
...
+
+        Range: [0, 1]
+        Units: dimensionless"""
+        limits = {
+            dictionary hard = {
+                float minimum = 0
+                float maximum = 1
+            }
+        }
+    )
```
andrewkaufman: @gdaviet similarly please review all of these material properties with the same scrutiny

tests/assets/sand.usda
```diff
@@ -0,0 +1,74 @@
+#usda 1.0
```
andrewkaufman, line 1: We don't typically provide these examples, but I think its useful & helped me understand your material design. So I'd like to keep it, just noting its non-standard thus far. Maybe we should have more for the other schemas in a followup PR.

newton_usd_schemas/generatedSchema.usda
andrewkaufman: I'm also wondering if it should be `NewtonParticleAPI` or `NewtonParticlesAPI` (since its `UsdGeomPoints`)

newton_usd_schemas/generatedSchema.usda
```diff
@@ -339,6 +339,217 @@ class "NewtonKaminoSceneAPI" (
...
+
+    uniform token newton:mpm:colliderVelocityMode = "forward" (
+        allowedTokens = ["forward", "backward"]
+        doc = """Method used to compute collider velocity.
+
+        "forward" uses the current body velocity. "backward" derives velocity
+        from the previous time-step position."""
+    )
+
+    uniform float newton:mpm:voxelSize = -inf (
```
andrewkaufman, line 373: Doc wording for the `-inf` sentinels (`voxelSize`, `airDrag`, `youngsModulus`, `yieldPressure`) drifts from the phrasing this repo already settled on.

`47351a8` ("Use engine-agnostic wording for -inf sentinel doc strings") standardized a single sentence:

> If `newton:sdfPadding` is set to `-inf`, the engine's default should be used instead.

This PR instead says "A value of `-inf` asks the engine to use its own default". Suggest matching the established sentence so the file stays uniform.

tests/test_scene_solver_mpm.py
```diff
@@ -0,0 +1,111 @@
...
+    def test_api_application_and_limitations(self):
+        self.assertTrue(self.scene.CanApplyAPI("NewtonMPMSceneAPI"))
+        self.scene.ApplyAPI("NewtonMPMSceneAPI")
+        self.assertTrue(self.scene.HasAPI("NewtonSceneAPI"))
+        self.assertTrue(self.scene.HasAPI("NewtonMPMSceneAPI"))
+
+        prim: Usd.Prim = self.stage.DefinePrim("/NotScene", "Xform")
+        self.assertFalse(prim.CanApplyAPI("NewtonMPMSceneAPI"))
+
+    def test_fallbacks_and_engine_default_sentinels(self):
```
andrewkaufman: The `limits` dictionaries this PR adds aren't covered by any test — 21 of them across the two new APIs (17 `hard`, 4 `soft`).

All 8 existing test files (`test_scene.py`, `test_scene_solver_xpbd.py`, `test_scene_solver_kamino.py`, `test_material.py`, `test_collision.py`, `test_joint.py`, `test_mass.py`, `test_actuator.py`) assert them via the `USD_HAS_LIMITS` / `GetHardLimits()` pattern, as part of the per-attribute tests described in my comment on `tests/test_mpm_material.py`. Please add the same coverage here.

This matters more than usual because `limits` metadata is hand-maintained in `generatedSchema.usda`, so an untested limit is the easiest thing in this repo to get silently wrong — the `poissonsRatio` doc-vs-`limits` mismatch in this PR is exactly that, and these assertions would have caught it.

tests/test_mpm_material.py
```diff
@@ -0,0 +1,67 @@
...
+        self.assertTrue(self.material.CanApplyAPI("NewtonMPMMaterialAPI"))
+        self.material.ApplyAPI("NewtonMPMMaterialAPI")
+        self.assertTrue(self.material.HasAPI("PhysicsMaterialAPI"))
+        self.assertTrue(self.material.HasAPI("NewtonMPMMaterialAPI"))
+        self.assertTrue(self.material.HasAttribute("physics:density"))
+
+        prim: Usd.Prim = self.stage.DefinePrim("/NotMaterial", "Xform")
+        self.assertFalse(prim.CanApplyAPI("NewtonMPMMaterialAPI"))
+
+    def test_fallbacks_and_engine_default_sentinels(self):
```
andrewkaufman: Please follow the existing per-attribute test style rather than a dict-driven loop.

Every existing test file uses one test per attribute — `test_soft_body_relaxation`, `test_joint_linear_relaxation`, `test_max_solver_iterations` and so on — each asserting the fallback value, a `Set()` round-trip, and the declared limits:

```python
def test_soft_body_relaxation(self):
    self.scene.ApplyAPI("NewtonXpbdSceneAPI")
    attr = self.scene.GetAttribute("newton:xpbd:softBodyRelaxation")
    self.assertIsNotNone(attr)
    self.assertAlmostEqual(attr.Get(), 0.9, places=5)

    success = attr.Set(0.8)
    self.assertTrue(success)
    self.assertAlmostEqual(attr.Get(), 0.8, places=5)

    if USD_HAS_LIMITS:
        hard = attr.GetHardLimits()
        self.assertTrue(hard.IsValid())
        self.assertAlmostEqual(hard.GetMinimum(), 0.0)
        self.assertAlmostEqual(hard.GetMaximum(), 1.0)
```

Collapsing the attributes into one `subTest` loop over a dict only checks the fallback value, so it drops the round-trip and limits coverage that the existing tests all have. The same applies to `test_fallbacks_and_engine_default_sentinels` and `test_allowed_tokens` in `tests/test_scene_solver_mpm.py`.

Naming also follows from this — existing files use `test_api_application` and `test_api_limitations` as separate tests, not a combined `test_api_application_and_limitations`.

tests/test_mpm_particle.py
```diff
@@ -0,0 +1,138 @@
...
+import pathlib
+import tempfile
+import unittest
+
+from pxr import Plug, Usd, UsdGeom, UsdPhysics, UsdShade, Vt
+
+import newton_usd_schemas  # noqa: F401
+
+
+def _assert_sand_stage(stage):
```
andrewkaufman: These helpers use bare `assert` (23 occurrences in this file). The rest of the suite uses `self.assertEqual` / `assertTrue` exclusively — currently **zero** bare asserts across `tests/`.

Beyond consistency, bare `assert` gives no diff on failure and vanishes under `python -O`.

Just define this function inline inside `test_sand_fixture_round_trip` since that is the only place its used

tests/assets/sand.usda
```diff
@@ -0,0 +1,74 @@
...
+
+def Xform "World"
+{
+    def PhysicsScene "PhysicsScene" (
+        prepend apiSchemas = ["NewtonMPMSceneAPI"]
+    )
+    {
+        float physics:gravityMagnitude = 9.81
+        vector3f physics:gravityDirection = (0, 0, -1)
+        int newton:maxSolverIterations = 100
```
andrewkaufman: This fixture is hand-authored and drifts from the form USD itself writes, in more than one way. Please round-trip it (`Usd.Stage.Open(...)` → `Export(...)`) and check in the normalized output.

What a round-trip currently changes:

- **`uniform` keyword omitted** where the schema declares it — `newton:maxSolverIterations`, `newton:mpm:rheologySolvers`, `newton:mpm:voxelSize`. The layer specs land as `VariabilityVarying` while the schema says `VariabilityUniform`. Values still resolve, since the schema wins at composition, but this is the one item with a real semantic mismatch behind it.
- **Scientific notation** — `1e7` / `1e5` / `2e7` / `2e5` become `10000000` / `100000` / `20000000` / `200000`.
- **`prepend apiSchemas`** becomes plain `apiSchemas`.
- **Property and metadata ordering**, plus `points` reflowed onto a single line.

Nothing here breaks the tests. It matters because this is the repo's first `.usda` asset, so it becomes the reference people copy when hand-authoring these schemas.

newton_usd_schemas/generatedSchema.usda
andrewkaufman: Backing this: `colliderBasis` and `strainBasis` are the only `newton:` token attributes in the repo with no `allowedTokens` (checked against the registry — every other one across the scene, material and collision APIs declares them). The practical cost is that `attr.Set("totally-not-a-basis")` succeeds silently and a UI has nothing to enumerate.

Splitting into a closed token set plus an integer parameter for the `pic` case would restore both validation and discoverability.

newton_usd_schemas/generatedSchema.usda
andrewkaufman: Same support here — splitting keeps `allowedTokens` closed so the attribute stays validatable and enumerable, which is otherwise lost for this one and `colliderBasis` alone.

tests/test_mpm_particle.py
```diff
@@ -0,0 +1,138 @@
...
+        bound_material, _relationship = subset_binding.ComputeBoundMaterial("physics")
+        self.assertEqual(bound_material.GetPath(), material.GetPath())
+
+    def test_sand_fixture_round_trip(self):
+        fixture = pathlib.Path(__file__).parent / "assets" / "sand.usda"
+        stage = Usd.Stage.Open(fixture.as_posix())
+        self.assertTrue(stage)
+        _assert_sand_stage(stage)
+
+        with tempfile.TemporaryDirectory() as directory:
```
andrewkaufman: What is this round-trip demonstrating?

newton_usd_schemas/generatedSchema.usda
gdaviet: Right, I think that makes sense. The short string identifiers used by the solver combine three things:
basisType: token ("linear" == "P", "trilinear" ==  "Q" , "bspline" == "B", "serendipity" == "S", "particle" == "pic")
basisOrder: int 
discontinuousBasis: boolean

Not all combinations are valid, but that what is implemented in the solver is prone to change, not validation should not necessarily done at Usd level
Wonder if there's a way to encode this basis "struct" as a way that can be reused for multiple attributes

newton_usd_schemas/generatedSchema.usda
gdaviet: Agree, I posted a similar comment on the original issue. 
But I wonder if we even need to encode all options in the schema, for instance I'll likely add an "apgd" option soon. Not sure what is the Usd precedent for these kind of attributes.

newton_usd_schemas/generatedSchema.usda
gdaviet: I am not entirely sure on what constitutes the bar for being considered stable at the Usd level,
so I've tried sorting the attributes into a few categories. I would probably include the first 3. 

Objective, highly unlikely to change:
 - tolerance
 - voxelSize

Attribute will survive, but allowed values may change over time:
 - rheologySolvers 
 - transferScheme
 - integrationScheme
 - velocityBasis
 - strainBasis
 - colliderBasis
 - gridType


Semantics may change over the next year:
 - criticalFraction 
 - airDrag 
 
 - maxActiveCellCount 
 - gridPadding



Internal details, can help make a scene work but should aim to remove:
 - colliderVelocityMode
 - colliderNormalFromSdfGradient
 - maxLeafNodeCount
 - maxUpperNodeCount
 - maxLowerNodeCount
 - warmstartMode

andrewkaufman: > Objective, highly unlikely to change

Definitely worth including

> Attribute will survive, but allowed values may change over time:

That is fine too, the `allowedTokens` and `limits` metadata are more like "ui hints" nothing actually enforces them aside from automated widgets (e.g. dropdowns & sliders) & you can still set values outside the ranges.

As long as the `type` is stable, we should include these in the schema & we can patch updates to allowed values as they change in the runtime.

> Semantics may change over the next year:

That's a long enough timeframe that I'd say we should include them & we can always re-word the tooltip and/or change the fallback value when they stabalize. Its technically a breaking change to change the fallback, but we're still a v0.x project & we can provide temporary backwards compatibility in newton parser _maybe_

> Internal details, can help make a scene work but should aim to remove:

I'd lean towards not include these ones then, unless @Milad-Rakhsha-NV @maxkra15 really insist they are essential for Isaac Sim workflows. But if that's the case it seems like you'd want to reconsider them belonging in one of the other categories anyways.

newton_usd_schemas/generatedSchema.usda
andrewkaufman: I think we should expose the `allowedTokens` that we _want to see in a dropdown_. That's what the field is realistcally used for. It could also be used for automated asset validation. But USD still allows you to set values outside the allowed tokens, so new values can get added later & people can still experiment with them before that happens.

newton_usd_schemas/generatedSchema.usda
andrewkaufman: I'd advocate making the schema less criptic and expand to 3 attrs per basis then:

```
uniform token newton:mpm:colliderBasisType = "linear"
   allowedTokens = ["linear", "trilinear", "bspline", "serendipity", "particle"]
uniform int newton:mpm:colliderBasisOrder = 2
uniform bool newton:mpm:colliderDiscontinuousBasis = false
```

I think its fine to just "replicate the pattern" rather than try to come up with some struct-like system (as USD doesnt have great options there). There's only two of them anyways (right?).

I don't mind if we go with `newton:mpm:colliderBasisType` vs `newton:mpm:collider:basisType`... Personally I don't feel the need for the extra midfix namespace, but if others disagree I'm not opposed to it.

The runtime parser can warn or error on invalid combinations, and once we have #10 we can add an asset validator that does the same.

newton_usd_schemas/generatedSchema.usda
Milad-Rakhsha-NV: @andrewkaufman do we need to define a physics width for particles to avoid colliding with the particle width that may be used for other things like rendering? Not a deal breaker, but I was thinking what happens if the user wants to have a different mpm particle width and UsdGeomPoints particle width?

newton_usd_schemas/generatedSchema.usda
Milad-Rakhsha-NV: I spelled out gs > Gause-Seidel, but I think we might want to do `cr > conjugate` residual and `cg > conjugate gradient`? but the for `GMRES > Generalized minimal residual` method, this is to long Or `APGD > Accelerated Projected Gradient Descent` . 
Not sure, but maybe we can keep gs as GS?  And just describe what they are in the attribute docs

newton_usd_schemas/generatedSchema.usda
andrewkaufman: We could. Curves deformables did that for normals & thickness iirc for similar reasoning.

That would make 3 widths attributes on the prim though & I think that gets pretty confusing... Note that `Points` `widths` and `primvars:widths` are already used directly in UsdPhysics (they drive scale/radius if you use a `PointsInstancer` with colliders).

Can we wait and add the extra `particle:widths` later on if it really comes up in practice that someone needs to separate from graphics widths?

newton_usd_schemas/generatedSchema.usda
```diff
@@ -866,6 +1053,143 @@ class NewtonMaterialAPI "NewtonMaterialAPI" (
     )
 }
 
+class NewtonMPMMaterialAPI "NewtonMPMMaterialAPI" (
+    apiSchemas = ["PhysicsMaterialAPI"]
```
andrewkaufman, line 1094: I missed this before, but its not applying `NewtonMaterialAPI` and I assume that is intentional? ie. the attribites on that API don't make sense for MPM?

maxkra15: Yes this is intentional. NewtonMaterialAPI describes surface-contact response, while this API provides MPM constitutive parameters. NewtonMPMMaterialAPI therefore applies only PhysicsMaterialAPI; the schema docs and a regression test now make that distinction explicit. users can still apply NewtonMaterialAPI separately when surface-contact properties are needed

tests/test_mpm_particle.py
maxkra15: removed it. fixture test now verifies schema application, authored values, simulation ownership, physics material bindings, and the point GeomSubset directly

newton_usd_schemas/generatedSchema.usda
andrewkaufman: Do you think we should consider renaming `NewtonMaterialAPI` to make that more clear? Or shuffling the contact specific attrs onto a `NewtonContactMaterialAPI`?

Are the coloumb friction paramters from `PhysicsMaterialAPI` actually useful for MPM or is that just a requirement because parsers ignore materials without that API as a marker?

newton_usd_schemas/generatedSchema.usda
maxkra15: Implemented this categorization. The objective, solver-selection, and currently useful configuration fields remain, but while warmstartMode, collider velocity/normal controls, and the leaf/lower/upper node capacity fields have been removed as runtime details

newton_usd_schemas/generatedSchema.usda
maxkra15: Agreed and implemented as the singular `NewtonParticleAPI`. It now uses the standard `physics:simulationOwner` relationship and reuses widths or `primvars:widths`; no third physical-width attribute added. The separate Newton importer branch handles indexed primvars and constant, vertex, and varying interpolation. We can add a dedicated physical width later

newton_usd_schemas/generatedSchema.usda
maxkra15: Reduced the docs to the schema’s required semantics, wording should also be more concise

newton_usd_schemas/generatedSchema.usda
maxkra15: Removed - omitted from the Newton MPM API and deferred to proposed PhysicsVolumeDeformableMaterialAPI

newton_usd_schemas/generatedSchema.usda
maxkra15: Removed - omitted from the Newton MPM API and deferred to proposed PhysicsVolumeDeformableMaterialAPI

newton_usd_schemas/generatedSchema.usda
maxkra15: Young modulus and Poisson ratio have been removed and deferred to the proposed PhysicsVolumeDeformableMaterialAPI. remaining fields match Newton’s current MPM constitutive model, with runtime matching types, defaults, units, and limits. @gdaviet, could you confirm this retained set-- elastic damping time, internal friction, yield pressure, tensile yield ratio, yield stress, viscosity, hardening, hardening/softening rates, and dilatancy, or flag anything that should remain runtime-only?

newton_usd_schemas/generatedSchema.usda
maxkra15: Standardized the remaining -inf sentinel documenttaion to the repository established wording. also
duplicated Young s modulus attribute was removed

tests/test_scene_solver_mpm.py
maxkra15: Added explicit per-attribute assertions for every remaining hard and soft limit including the sentinel-backed soft-limit cases

tests/test_mpm_material.py
maxkra15: Refactored the tests to match the existing suite: one test per attribute covering its fallback, Set() behavior, and declared limits or allowed tokens. API application and limitations are now separate tests

tests/test_mpm_particle.py
maxkra15: Removed the generic helper and test, no bare assertions remain. The semantic fixture test now uses unittest assertions directly.

tests/assets/sand.usda
maxkra15: Normalized sand.usda through USD export semantics. now uses schema-declared variability, normalized numeric formatting, API metadata, and property ordering

newton_usd_schemas/generatedSchema.usda
maxkra15: removed the redundant gs* aliases and retained only the gauss-seidel* spellings. I kept cg, cr, and gmres as the conventional algorithm names used by the runtime

newton_usd_schemas/generatedSchema.usda
maxkra15: Implemented the attribute pattern for both collider and strain bases: type, order, and discontinuity. The type attributes now have enumerated allowedTokens. For the particle family, order and discontinuity are ignored and point capacity remains engine-defined. If picN capacity needs to be serialized, it could be a dedicated attribute rather than overloading basis order

newton_usd_schemas/generatedSchema.usda
Milad-Rakhsha-NV: I think we shouldn't separate contact-related ones into NewtonContactMaterialAPI. I believe physicsMaterialAPI is mostly used for rigid body contact, and the interpretation is somewhat different from continuum material interpretation

gdaviet: Conceptually there is a relationship between the contact properties of particles (NewtonMaterialAPI) and the constitutive model of MPM (PhysicsMaterialAPI), so if we really wanted too we could reuse the friction, contact elasticity and contact damping for the MPM friction, young modulus and elastic damping. I don't believe that would make scenes easier to author though.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -866,6 +1075,146 @@ class NewtonMaterialAPI "NewtonMaterialAPI" (
...
+    `NewtonMaterialAPI`, which describes surface-contact response. Use
+    `PhysicsVolumeDeformableMaterialAPI` for Young's modulus and Poisson's
+    ratio."""
+)
+{
+    float newton:mpm:elasticDampingTime = 0 (
+        doc = """Elastic damping relaxation time.
+
+        Range: [0, inf)
+        Units: time"""
```
gdaviet: Newton has now moved to absolute damping everywhere, so this attribute should probably be exposed as absolute as well. Unit would be [Pa.s] or (or force x time / (displacement x displacement)), and the loaded would need to divide by the youngModulus until the MPM solver is updated to directly accept absolute values (I was sure I had done that already, but apparently not)

newton_usd_schemas/generatedSchema.usda
```diff
@@ -339,6 +339,196 @@ class "NewtonKaminoSceneAPI" (
...
+
+    uniform bool newton:mpm:colliderDiscontinuousBasis = false (
+        doc = """Whether the collider basis is discontinuous between grid cells.
+
+        Ignored for the "particle" family and at order zero, where the basis is
+        inherently discontinuous."""
+    )
+
+    uniform token newton:mpm:strainBasisType = "linear" (
+        allowedTokens = ["linear", "trilinear", "bspline", "serendipity", "particle"]
```
gdaviet: If we are redefining the accepted token list for each kind of basis then we can remove `bspline`, `serendipity` from strainBasisType, they will be rejected by the solver

newton_usd_schemas/generatedSchema.usda
```diff
@@ -339,6 +339,196 @@ class "NewtonKaminoSceneAPI" (
...
+    )
+
+    uniform bool newton:mpm:strainDiscontinuousBasis = false (
+        doc = """Whether the strain basis is discontinuous between grid cells.
+
+        Ignored for the "particle" family and at order zero, where the basis is
+        inherently discontinuous."""
+    )
+
+    uniform token newton:mpm:velocityBasis = "Q1" (
```
gdaviet: We could also have the velocity basis follow a similar pattern. Accepted basis types: `trilinear`, `bspline`. Orders 1, 2, 3. No discontinuous flag for this one, velocities must be continuous

newton_usd_schemas/generatedSchema.usda
```diff
@@ -866,6 +1075,146 @@ class NewtonMaterialAPI "NewtonMaterialAPI" (
...
+            dictionary hard = {
+                float minimum = 0
+            }
+        }
+    )
+
+    float newton:mpm:internalFriction = 0.5 (
+        doc = """Pressure-dependent internal friction coefficient.
+
+        This controls the MPM yield response rather than surface contact.
```
gdaviet: Would reformulate as: ```
Internal friction coefficient.

        This controls the MPM pressure-dependent yield response rather than surface contact.
```

File: newton_usd_schemas/generatedSchema.usda [hunk 25]
```diff
@@ -866,6 +1075,146 @@ class NewtonMaterialAPI "NewtonMaterialAPI" (
...
+        limits = {
+            dictionary hard = {
+                float minimum = 0
+                float maximum = 1
+            }
+        }
+    )
+
+    float newton:mpm:yieldStress = 0 (
+        doc = """Deviatoric yield stress.
```
gdaviet: Pressure-independent deviatoric yield stress.

Added on top of the pressure-dependent yield-stress defined by newton:mpm:internalFriction.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -866,6 +1075,146 @@ class NewtonMaterialAPI "NewtonMaterialAPI" (
...
+        Units: force * time / (distance * distance)"""
+        limits = {
+            dictionary hard = {
+                float minimum = 0
+            }
+        }
+    )
+
+    float newton:mpm:hardening = 0 (
+        doc = """Isotropic hardening coefficient applied to the yield surface.
```
gdaviet, line 1201: May add:

Defines how quickly the yield surface expands and shrinks with the plastic volume. 
A value of 0 disables the hardening model.

gdaviet (commented):
I am realizing that on missing attribute is `particle_Jp`, the initial plastic volume strain. This is a state attribute on the MPM solver, but could be part of the MPM material as well. Defining an initialPlasticVolumeStrain is important when using an hardening law to prevent collapse at the start of the simulation.

newton_usd_schemas/generatedSchema.usda
maxkra15: Agreed — keeping them separate.

newton_usd_schemas/generatedSchema.usda
maxkra15: Done — absolute `elasticDamping` [Pa·s]; loader conversion in the Newton parser PR.

newton_usd_schemas/generatedSchema.usda
maxkra15: Done.

newton_usd_schemas/generatedSchema.usda
maxkra15: Done.

newton_usd_schemas/generatedSchema.usda
maxkra15: Done.

newton_usd_schemas/generatedSchema.usda
maxkra15: Done.

newton_usd_schemas/generatedSchema.usda
maxkra15: Done.

maxkra15:
Done — `newton:mpm:initialPlasticVolumeStrain`, default `1`.

newton_usd_schemas/generatedSchema.usda
Milad-Rakhsha-NV: hmm I don't think we should follow the runtime names for schema. 
```
"gs", "gauss-seidel", "gs-soa", "gauss-seidel-soa", "gs-batched", "gauss-seidel-batched", "jacobi", "cg", "cr", "gmres"
```
I don't fully understand why we should have "gs" AND "gauss-seidel" as allowed tokens here. 
If we really want to use acronyms, we should be consistent across all tokens and use uppercase as well IMO. The mapping from these to actual runtime names should be the job of the parser I believe and there is no need to make the schemas cryptic

chschuma-disney (dismissed):
The schema looks good to me now.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -866,6 +1090,168 @@ class NewtonMaterialAPI "NewtonMaterialAPI" (
     )
 }
 
+class NewtonMPMMaterialAPI "NewtonMPMMaterialAPI" (
+    apiSchemas = ["PhysicsMaterialAPI"]
+    doc = """Provides constitutive parameters for an implicit MPM solver.
+
+    Applying this schema also applies `PhysicsMaterialAPI`. It does not apply
+    `NewtonMaterialAPI`, which describes surface-contact response. Use
+    `PhysicsVolumeDeformableMaterialAPI` for Young's modulus and Poisson's
```
andrewkaufman, line 1099: We should double check with @sschirm where the YM & PS landed in the deformables propsoal. I think it was being re-worked to a general `PhysicsDeformableMaterialAPI`

newton_usd_schemas/generatedSchema.usda
```diff
@@ -339,6 +339,211 @@ class "NewtonKaminoSceneAPI" (
...
+            }
+        }
+    )
+
+    uniform token[] newton:mpm:rheologySolvers = ["auto"] (
+        allowedTokens = ["auto", "gauss-seidel", "gauss-seidel-soa", "gauss-seidel-batched", "jacobi", "cg", "cr", "gmres"]
+        doc = """Ordered rheology solver sequence.
+
+        A sequence with more than one entry warm-starts solvers from left to
+        right. The "auto" mode selects an implementation based on the velocity
```
andrewkaufman, line 369: Is it valid to supply `["auto", ...]` or is auto mode only valid as the only solver choice? What happens when someone provides `auto` as the Nth solver rather than the first?

newton_usd_schemas/generatedSchema.usda
andrewkaufman: My appologies for flip-flopping here, but I confirmed that while `PhysicsVolumeDeformableMaterialAPI` is the correct API name currently, it is being re-named, AND more importantly its only applicable to `UsdGeom.TetMesh` prims. There is no base class in the proposal from which to inherit Young's modulus and Poisson's ratio.

Instead, we should add those back here under newton namespace, but match the naming, units, and range of the proposal (updated in https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/111/commits/61d83b54b7efbe97ad2f480de885255cd1e593be) so that we can more easily propose back the Particles extension later on.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -765,6 +970,25 @@ class NewtonSiteAPI "NewtonSiteAPI" (
 {
 }
 
+class NewtonParticleAPI "NewtonParticleAPI" (
+    doc = """Applies physics properties to a `UsdGeomPoints` prim and marks it as
+    driven by a simulation.
+
+    Particle positions, velocities, widths, and stable identifiers use the
+    standard `UsdGeomPoints` `points`, `velocities`, `widths` (or
+    `primvars:widths`), and `ids` attributes. Widths are diameters, so radius is
```
andrewkaufman, line 979: The updated AOUSD proposal https://github.com/PixarAnimationStudios/OpenUSD-proposals/pull/111/commits/61d83b54b7efbe97ad2f480de885255cd1e593be does add simulation thickness attributes to the other deformable types, but on the material not the geometry prim. eg. `physics:surfaceThickness`. We could do that here too `newton:particleThickness` or something... though tbh `radius` seems more natural :shrug: 

If you're ok I think we can still leave this out for now & add it in a followup perhaps. Just wanted to note it incase you were really wanting a separation from the graphics widths.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -765,6 +970,25 @@ class NewtonSiteAPI "NewtonSiteAPI" (
 {
 }
 
+class NewtonParticleAPI "NewtonParticleAPI" (
```
andrewkaufman, line 973: I start to wonder if this should instead be `NewtonPointsDeformableSimAPI` to match the naming of the AOUSD proposal... while I find that _very redundant_ it does sit nicely alongside the others (`PhysicsVolumeDeformableSimAPI`, `PhysicsSurfaceDeformableSimAPI`, `PhysicsCurvesDeformableSimAPI`)

and then makes it easier for us to justify the per point `float[] physics:masses ()` that they all have

> The simulation APIs provide a `masses` attribute for specifying a per-point mass on the simulation geometry. When the `masses` attribute is specified, the size of the array must match the number of points defined in the simulation geometry's `points` attribute. These per-point mass definitions take precedence over any mass or density parameters specified by `UsdPhysicsDeformableBodyAPI` or `UsdPhysicsMaterialAPI`.

as well as rest points if we want that

> `point3f[] physics:restShapePoints ()`

if we do all this, the sand.usda example should be updated with both `PhysicsDeformableBodyAPI` and `NewtonPointsDeformableSimAPI` applied (note `PhysicsDeformableBodyAPI` can be on a parent of the actual `Points` prim, while `NewtonPointsDeformableSimAPI` would need to be on the `Points` directly.

andrewkaufman (commented):
Sorry @maxkra15 I need to flip-flop my request about YM & PS... The AOUSD proposal updated today with more details. Comments are inline.

newton_usd_schemas/generatedSchema.usda
andrewkaufman: If we do this, `PhysicsDeformableBodyAPI` does already provide us the `simulationOwner` relationship, so we don't need the newton one.

newton_usd_schemas/generatedSchema.usda
maxkra15: done, restored newton:mpm:youngsModulus and newton:mpm:poissonsRatio with the AOUSD spelling, units, and ranges, for  young’s modulus retain Newton s -inf engine-default sentinel

newton_usd_schemas/generatedSchema.usda
maxkra15: auto is valid at any position; each occurrence resolves in place from the velocity basis before the ordered warm-start chain runs. Clarified and tested with ["cg", "auto"]

newton_usd_schemas/generatedSchema.usda
maxkra15: leaving this out. widths remains the physical diameter; a separate material radius can be added if a graphics/simulation split is needed

newton_usd_schemas/generatedSchema.usda
maxkra15: renamed to NewtonPointsDeformableSimAPI, added physics:masses, and updated sand.usda to apply PhysicsDeformableBodyAPI. omitted restShapePoints because Newton MPM has no compatible rest-point semantics

maxkra15: Also removed physics:simulationOwner from NewtonPointsDeformableSimAPI; ownership now comes from PhysicsDeformableBodyAPI

newton_usd_schemas/generatedSchema.usda
andrewkaufman: I think this conversation is resolved on the latest code

```
        allowedTokens = ["auto", "gauss-seidel", "gauss-seidel-soa", "gauss-seidel-batched", "jacobi", "cg", "cr", "gmres"]
```

andrewkaufman (approved):

newton_usd_schemas/generatedSchema.usda
Milad-Rakhsha: Shouldn't we expand `cr` to conjugate residual and `cg` to conjugate gradient as well? 
I leave the decision to @gdaviet and @andrewkaufman but I think for consistency we should

chschuma-disney (approved):

PR #89, opened 2026-09-08
Add NewtonCollisionAPI 

gyeomannvidia:
## Description

Closes https://github.com/newton-physics/newton-usd-schemas/pull/89

## Checklist


newton_usd_schemas/_version.py
```diff
@@ -1,4 +1,4 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 The Newton Developers
 # SPDX-License-Identifier: Apache-2.0
 
-__version__ = "0.5.0"
+__version__ = "0.6.0.dev0"
```
gyeomannvidia, line 4: This needs to be removed before merging.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -544,6 +544,238 @@ class "NewtonMPMSceneAPI" (
...
+        Range: [-1, inf)
+        Units: dimensionless"""
+        limits = {
+            dictionary hard = {
+                int minimum = -1
+            }
+        }
+    )
+
+    uniform float newton:collisionPipeline:softContactGap = -inf (
```
mzamoramora-nvidia, line 618: Since `-inf` is the documented sentinel, could we use a soft minimum of `0` here instead of a hard minimum? That would keep the non-negative authoring guidance while allowing the sentinel. The same applies to `speculativeMaxExtension`.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -544,6 +544,238 @@ class "NewtonMPMSceneAPI" (
...
+
+    uniform bool newton:collisionPipeline:includeStaticKinematicPairs = true (
+        doc = """Whether to generate contacts for shape pairs where both shapes
+        are immovable.
+
+        Set to false to filter static-static, static-kinematic, and
+        kinematic-kinematic pairs."""
+    )
+
+    uniform int newton:collisionPipeline:shapePairsMax = -1 (
```
mzamoramora-nvidia, line 670: Could we make the schema and Newton agree on whether `0` is valid here? The schema currently allows it, but `CollisionPipeline` requires a positive value or `None`, so a schema-valid stage with `0` fails to load.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -544,6 +544,238 @@ class "NewtonMPMSceneAPI" (
...
+        mode."""
+    )
+
+    uniform bool newton:collisionPipeline:verifyBuffers = true (
+        doc = """Run a diagnostic kernel at the end of the narrow phase that
+        prints warnings on any intermediate candidate-pair or final rigid
+        contact buffer overflow."""
+    )
+
+    uniform float newton:collisionPipeline:contactReductionHashtableSizeFactor = 0.25 (
```
mzamoramora-nvidia, line 739: The description and Newton validation require this value to be greater than zero, while the hard minimum currently allows zero. Could we align the schema metadata with the runtime check?

newton_usd_schemas/generatedSchema.usda
gyeomannvidia: Nice catch.  I updated the schema doc to make clear that 0 is invalid and added an error message and a test of the error message in newton.

newton_usd_schemas/generatedSchema.usda
gyeomannvidia: I'm not sure this has a solution for floating point numbers. 
The problem is replicated multiple times eg newton:hydroelasticStiffness, newton:sdfPadding.

mzamoramora-nvidia: You’re right. The Newton side already [enforces the strict `> 0` check](https://github.com/newton-physics/newton/blob/4bf7a4ad2e130e302403fd60d10e4a9837bed41a/newton/_src/sim/collide.py#L2049-L2057), and [commit `00e1e57`](https://github.com/newton-physics/newton/commit/00e1e57cbacbbe799726affbc74b02dea4a98dbf) adds coverage for zero, so I think the current handling should be fine. Thanks!

newton_usd_schemas/generatedSchema.usda
```diff
@@ -544,6 +544,235 @@ class "NewtonMPMSceneAPI" (
     )
 }
 
+class "NewtonCollisionPipelineAPI" (
+    apiSchemas = ["NewtonSceneAPI"]
+    doc = """Configures Newton's collision pipeline (broad phase, narrow phase
```
jcarius-nv, line 549: **P1 — Define this as an experimental runtime/application profile.** Bundling the first version is reasonable while we gain authoring experience, but this API mixes durable contact intent with machine/workload policy such as buffer capacities, diagnostics, autodiff, and broad-phase selection. Please state that distinction here so users do not mistake every attribute for portable physical asset semantics, and so we retain room to split the profile after real workflow evidence.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -544,6 +544,235 @@ class "NewtonMPMSceneAPI" (
...
+            dictionary hard = {
+                int minimum = 1
+            }
+        }
+    )
+
+    uniform int newton:collisionPipeline:rigidContactMax = -1 (
+        doc = """Maximum rigid contact buffer capacity.
+
+        If `newton:collisionPipeline:rigidContactMax` is set to `-1`, the
```
jcarius-nv, line 584: **P2 — Describe the actual `-1` resolution order.** Newton omits this argument for `-1`, after which `CollisionPipeline.__init__` first preserves a positive `model.rigid_contact_max` and only estimates when the model value is absent/non-positive. Saying that `-1` always estimates is therefore inaccurate; referring to the constructor's normal model-or-estimate resolution would match runtime behavior.

jcarius-nv (commented):
I am comfortable keeping the authoring surface bundled for this first version if it is explicitly framed as experimental runtime configuration. Please also rename the PR to `Add NewtonCollisionPipelineAPI` and replace the self-closing template description with a standalone statement of scope and intended workflow. I have not repeated the existing version or limit-metadata threads.

newton_usd_schemas/generatedSchema.usda
```diff
@@ -544,6 +544,235 @@ class "NewtonMPMSceneAPI" (
...
+        allowedTokens = ["nxn", "sap", "explicit"]
+        doc = """Broad phase candidate-pair generation strategy.
+
+        "nxn": all-pairs broad phase with AABB overlap, group, and excluded-pair
+        filtering. "sap": sweep-and-prune broad phase. "explicit": use a
+        precomputed shape-pair list (e.g. `PhysicsFilteredPairsAPI`-derived
+        pairs) instead of dynamic broad-phase search."""
+    )
+
+    uniform int newton:collisionPipeline:maxTrianglePairs = 1000000 (
```
chschuma-disney: We could consider settings this to a `-inf` sentinel to signal that the simulator can choose its own default, so we can update Newton's default without any changes to the schema.

